PR #5, opened 2024-07-08
Job scheduler service

adam-fowler:
New Service to schedule jobs at specific times
```swift
let jobSchedule = JobSchedule([
    (job: SundayCleanupJob(), schedule: .weekly(day: .sunday, hour: 4)),
    (job: BirthdayJob(), schedule: .yearly(month: .march, date: 4)),
])
let serviceGroup = ServiceGroup(
    configuration: .init(
        services: [jobQueue, jobSchedule.scheduler(on: jobQueue)]
    )
)
```


Sources/Jobs/Scheduler/Schedule.swift
```diff
@@ -0,0 +1,155 @@
...
+        case specific(Value)
+    }
+
+    let second: Parameter<Int>
+    let minute: Parameter<Int>
+    let hour: Parameter<Int>
+    let date: Parameter<Int>
+    let month: Parameter<Month>
+    let day: Parameter<Day>
+    let calendar: Calendar
```
adam-fowler, line 58: Need to sort Sendable conformance on Linux

Sources/Jobs/Scheduler/JobSchedule.swift
```diff
@@ -0,0 +1,189 @@
...
+            }
+
+            mutating func next() async -> JobParameters? {
+                guard let job = self.jobSchedule.nextJob() else {
+                    return nil
+                }
+                self.logger.debug(
+                    "Next Scheduled Job",
+                    metadata: [
+                        "_job_type": .stringConvertible(type(of: job.element.jobParameters).jobName),
```
thoven87, line 123: Is this the name of a job?

adam-fowler: Renamed it to JobName

adam-fowler:
Added stored metadata to queue and using this to store last date job was queued. 

Also added an accuracy parameter to job schedules which indicates how accurately the schedule should be followed if we fall behind.

Tests/JobsTests/JobSchedulerTests.swift
```diff
@@ -152,7 +152,7 @@ final class JobSchedulerTests: XCTestCase {
         jobSchedule.addJob(TriggerShutdownParameters(), schedule: .everyMinute(second: dateComponents.second!))
 
         await withThrowingTaskGroup(of: Void.self) { group in
-            let serviceGroup = ServiceGroup(
+            let serviceGroup = await ServiceGroup(
```
thoven87, line 155: Is it possible that `setMetada` or `getMetada` could be called before a persist driver is ready?

adam-fowler: We aren't using the persist driver as it would pull in the Hummingbird repository as a dependency. I guess you mean before the Postgres client has initialised? I have a fix for this, in hb-postgres. It is a bit of a hack, but basically I wait 100 microseconds to ensure the Postgres client has started up.

thoven87: Yes, I meant before the Postgres client has initialized. `It is a bit of a hack, but basically I wait 100 microseconds to ensure the Postgres client has started up` This could be flaky, should this where application might need restart incase the Postgres client takes more 100 microseconds to initialize?

adam-fowler: The thing is queries don't fail. It's just the connections manager isn't running. It's a pain because the error is basically false in this situation because the connnection manager has been started it just hasn't got to the point where it has set the flag saying it is running.

thoven87: > The thing is queries don't fail. It's just the connections manager isn't running. It's a pain because the error is basically false in this situation because the connnection manager has been started it just hasn't got to the point where it has set the flag saying it is running.

I see! So, migrations would still apply without any issues and mediate would get saved?

adam-fowler: > see! So, migrations would still apply without any issues and mediate would get saved?

Yes

thoven87: Thank You! I don't have anymore questions on this PR. Also, please mark this conversion as resolved. I don't any options mark as resolved on my end.

Joannis (changes requested):
Only 1 change needed IMO, I think we're good to merge after that

Sources/Jobs/MemoryJobQueue.swift
```diff
@@ -60,16 +60,26 @@ public final class MemoryQueue: JobQueueDriver {
         }
     }
 
+    public func getMetadata(_ key: String) async -> ByteBuffer? {
+        await self.queue.getMetadata(key)
+    }
+
+    public func setMetadata(key: String, value: NIOCore.ByteBuffer) async {
```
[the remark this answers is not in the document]
Joannis: Nit

Joannis (approved):

PR #8, opened 2024-08-13
Don't pop job off queue until child task has finished

adam-fowler:
Previously we were popping job off queue, waiting on a previous child task to finish before running it. This change waits for the child task first.

### DIFF
--- a/Sources/Jobs/JobQueueHandler.swift
+++ b/Sources/Jobs/JobQueueHandler.swift
@@ -46,8 +46,9 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Service {
                         }
                     }
                 }
-                while let job = try await iterator.next() {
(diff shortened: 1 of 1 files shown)


Sources/Jobs/JobQueueHandler.swift
```diff
@@ -94,7 +95,7 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Service {
                     try await self.queue.failed(jobId: queuedJob.id, error: error)
                     return
                 } catch {
-                    if count == 0 {
+                    if count <= 0 {
```
Joannis, line 98: Would this ever be `-1`?

Joannis: Still, I appreciate safe programming

PR #10, opened 2024-08-15
always use fulfillment in tests

adam-fowler:


Tests/JobsTests/JobsTests.swift
```diff
@@ -283,7 +275,7 @@ final class JobsTests: XCTestCase {
                 for i in 0..<200 {
                     try await jobQueue.push(id: jobIdentifer, parameters: i)
                 }
-                await self.wait(for: [expectation], timeout: 5)
+                await fulfillment(of: [expectation], timeout: 5)
```
Joannis, line 278: Isn't this Swift 6 only? I remember some issues on Linux.

Joannis (approved):
If the tests succeed, LGTM

Tests/JobsTests/JobsTests.swift
adam-fowler: No since 5.9. I thought it was 5.10 but in actually fact works fine in 5.9. See CI results

PR #11, opened 2024-08-19
feat(queuedAt): Added when a job was queued at

thoven87:
This is a follow up to the previous PR which adds when a job was queued at. 

We can now plot average wait time from queued to processing.

### DIFF
--- a/Sources/Jobs/JobQueueHandler.swift
+++ b/Sources/Jobs/JobQueueHandler.swift
@@ -75,7 +75,7 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Service {
         Timer(
             label: "\(self.metricsLabel)_queued_for_duration_seconds",
             preferredDisplayUnit: .seconds
-        ).recordNanoseconds(Int64(jobQueuedDuration * 100_000))
+        ).recordNanoseconds(Int64(jobQueuedDuration * 1_000_000))
 
         Meter(label: self.meterLabel, dimensions: [("status", JobStatus.queued.rawValue)]).decrement()
         Meter(label: self.meterLabel, dimensions: [("status", JobStatus.processing.rawValue)]).increment()
--- a/Sources/Jobs/JobQueueHandler.swift
+++ b/Sources/Jobs/JobQueueHandler.swift
@@ -68,6 +68,12 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Service {
         var logger = logger
         let startTime = DispatchTime.now().uptimeNanoseconds
 
+        // Calculate wait time from queued to processing
+        Timer(
+            label: "\(self.metricsLabel)_queued_for_duration_seconds",
+            preferredDisplayUnit: .seconds
+        ).recordNanoseconds(startTime - queuedJob.queuedAt)
+
         Meter(label: self.meterLabel, dimensions: [("status", JobStatus.queued.rawValue)]).decrement()
         Meter(label: self.meterLabel, dimensions: [("status", JobStatus.processing.rawValue)]).increment()
         defer {
--- a/Sources/Jobs/JobQueueHandler.swift
+++ b/Sources/Jobs/JobQueueHandler.swift
@@ -75,7 +75,7 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Service {
         Timer(
             label: "\(self.metricsLabel)_queued_for_duration_seconds",
             preferredDisplayUnit: .seconds
-        ).recordNanoseconds(Int64(jobQueuedDuration * 1_000_000_000))
+        ).recordSeconds(jobQueuedDuration)
 
         Meter(label: self.meterLabel, dimensions: [("status", JobStatus.queued.rawValue)]).decrement()
         Meter(label: self.meterLabel, dimensions: [("status", JobStatus.processing.rawValue)]).increment()
--- a/Sources/Jobs/JobQueueHandler.swift
+++ b/Sources/Jobs/JobQueueHandler.swift
@@ -11,7 +11,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
-
+import struct Foundation.Date
 import Dispatch
 import Logging
 import Metrics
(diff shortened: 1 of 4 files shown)


Sources/Jobs/JobQueueHandler.swift
```diff
@@ -68,6 +68,12 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Service {
         var logger = logger
         let startTime = DispatchTime.now().uptimeNanoseconds
 
+        // Calculate wait time from queued to processing
+        Timer(
```
adam-fowler, line 72: This doesn't work. It assumes the machine running the job was the same as the machine that queued it. You need to use a value that has the same base eg 1 Jan 1970, not boot time.

adam-fowler (changes requested):
One issue, but otherwise looks fine
You need to also to create PRs in parallel for swift-jobs-redis (although this may just work out the box), and hummingbird-postgres jobs.

Sources/Jobs/JobQueueHandler.swift
Joannis: Exactly, this won't work if the job runs on another container instance or host

thoven87: Understood! The issue I faced is that, the swift date object is confusing to me. There's `timeIntervalSince1970`, `timeIntervalSinceNow`, `timeIntervalSinceReferenceDate` which one is available when or where is not clear.

thoven87:
> One issue, but otherwise looks fine You need to also to create PRs in parallel for swift-jobs-redis (although this may just work out the box), and hummingbird-postgres jobs.

I will follow up with other PRs for both swift-jobs-redis and hummingbird-postgres. Although the current implementation is implicit.

Sources/Jobs/JobQueueHandler.swift
thoven87: I updated the code to use Date.

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -68,6 +69,14 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Service {
         var logger = logger
         let startTime = DispatchTime.now().uptimeNanoseconds
 
+        let jobQueuedDuration = Date().timeIntervalSince(queuedJob.queuedAt)
+
+        // Calculate wait time from queued to processing
+        Timer(
+            label: "\(self.metricsLabel)_queued_for_duration_seconds",
+            preferredDisplayUnit: .seconds
+        ).recordNanoseconds(Int64(jobQueuedDuration * 100_000))
```
Joannis, line 78: This doesn't seem like a correct multiplication. Isn't a second `1_000_000` nanoseconds?

thoven87: Yep, that was a typo. Thanks!

adam-fowler: A second is 1_000_000_000 nanoseconds.

Joannis: You're right, 9 digits. Thanks for the correction

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -68,6 +69,14 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Service {
         var logger = logger
         let startTime = DispatchTime.now().uptimeNanoseconds
 
+        let jobQueuedDuration = Date().timeIntervalSince(queuedJob.queuedAt)
+
+        // Calculate wait time from queued to processing
+        Timer(
+            label: "\(self.metricsLabel)_queued_for_duration_seconds",
+            preferredDisplayUnit: .seconds
+        ).recordNanoseconds(Int64(jobQueuedDuration * 1_000_000))
```
Joannis, line 78:
```suggestion
        ).recordNanoseconds(Int64(jobQueuedDuration * 1_000_000_000))
```

Joannis (changes requested):

Joannis (approved):

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -68,6 +69,14 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Service {
         var logger = logger
         let startTime = DispatchTime.now().uptimeNanoseconds
 
+        let jobQueuedDuration = Date().timeIntervalSince(queuedJob.queuedAt)
```
adam-fowler, line 72:
```suggestion
        let jobQueuedDuration = Date.now.timeIntervalSince(queuedJob.queuedAt)
```

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -68,6 +69,14 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Service {
...
+
+        // Calculate wait time from queued to processing
+        Timer(
+            label: "\(self.metricsLabel)_queued_for_duration_seconds",
+            preferredDisplayUnit: .seconds
+        ).recordNanoseconds(Int64(jobQueuedDuration * 1_000_000_000))
```
adam-fowler, line 78: You are better using the `recordSeconds` function that takes a Double it deals with overflow.

I'm not a great fan of using Doubles as accuracy starts to fail in this situation after about 3 years, but it'd be quite a while longer before the accuracy is so far out it'd be noticeable.

adam-fowler (changes requested):

adam-fowler (approved):
We're good.

Before I merge this though I'd like to see PRs on swift-jobs-redis and hummingbird-postgres, using this branch (we can update them to use the released version once this is in a release). Very often when updating the implementations that is when I see issues, it also a good way to see how these changes impact the implementations

thoven87:
> We're good.
> 
> Before I merge this though I'd like to see PRs on swift-jobs-redis and hummingbird-postgres, using this branch (we can update them to use the released version once this is in a release). Very often when updating the implementations that is when I see issues, it also a good way to see how these changes impact the implementations

Thanks, Adam! I will open a Pull request in [Hummingbird Postgres](https://github.com/hummingbird-project/hummingbird-postgres) followed by [Swift Jobs Redis](https://github.com/hummingbird-project/swift-jobs-redis)

Joannis:
Looks good to merge, right @adam-fowler ?

Sources/Jobs/QueuedJob.swift
```diff
@@ -21,10 +21,13 @@ public struct QueuedJob<JobID: Sendable>: Sendable {
     public let id: JobID
     /// Job data
     public let jobBuffer: ByteBuffer
+    /// Job queued at
+    public let queuedAt: Date
 
     /// Initialize a queue job
-    public init(id: JobID, jobBuffer: ByteBuffer) {
+    public init(id: JobID, jobBuffer: ByteBuffer, queuedAt: Date = .init()) {
```
adam-fowler, line 28: queuedAt should not have a default value. It means we are missing places where we aren't setting the value correctly

thoven87: I set a default here because, I don't think we have any use for this field in the Redis Driver implementation. I will drop the default then.

adam-fowler (approved):

PR #14, opened 2024-08-23
Job coding re-organisation

adam-fowler:
- Move encoded job type to EncodableJob.swift
- Move decoded job type to AnyDecodableJob.swift
-Added `JobInstanceData` to hold any data related to the job
- Also added queuedAt time to `JobInstanceData`

And metrics recording as in #11

@thoven87


Sources/Jobs/EncodableJob.swift
```diff
@@ -0,0 +1,34 @@
...
+// See hummingbird/CONTRIBUTORS.txt for the list of Hummingbird authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+/// Type used to encode a job
+struct EncodabeJob<Parameters: Codable & Sendable>: Encodable, Sendable {
```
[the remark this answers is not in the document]
adam-fowler: internal, I've fixed the inits

Sources/Jobs/JobInstance.swift
```diff
@@ -0,0 +1,77 @@
...
+    var maxRetryCount: Int { get }
+    /// Time job was queued
+    var queuedAt: Date { get }
+    /// Function to execute the job
+    func execute(context: JobContext) async throws
+}
+
+extension JobInstanceProtocol {
+    /// name of job type
+    public var name: String {
```
thoven87, line 33: Should the JobName move out of Parameters? Right now, only one job can be registered at a time.

thoven87:
Just one small comment. Thanks for putting this together @adam-fowler.

Sources/Jobs/JobInstance.swift
adam-fowler: You can register multiple jobs to a queue handler. If you can't that's a bug. 

The JobParameters holds both the job parameters and the job identifier. The job name is a string version of the identifier that's used in encoding and decoding.

thoven87: Makes sense.

We can already do the the following today

```swift
queue.registerJob(
            parameters: EmailParameters.self,
            maxRetryCount: 10)
        { (parameters: EmailParameters, _) in
            do {
                try await emailJob.send(emailParameters: parameters)
            } catch {
                logger.info("Failed to send semail: \(error.localizedDescription)", metadata: [
                    "error": "\(String(reflecting: error))",
                ])
                throw error
            }
        }

queue.registerJob(
            parameters: StatsParameters.self,
            maxRetryCount: 5)
        { (parameters: StatsParameters, _) in
            do {
                try await statsJob.compute(date: Date.now)
            } catch {
                logger.info("Failed to compute stats: \(error.localizedDescription)", metadata: [
                    "error": "\(String(reflecting: error))",
                ])
                throw error
            }
        }

queue.registerJob(
            parameters: AnotherParameters.self,
            maxRetryCount: 4)
        { (parameters: AnotherParameters, _) in
        }
```

I was asking if the interface could change were the we pass the name of the job at the time of registration instead.

```swift
queue.registerJob(
            parameters: AnotherParameters.self,
            jobName: "SomeJobName"
            maxRetryCount: 4)
        { (parameters: AnotherParameters, _) in
        }
```

adam-fowler: Can we discuss this in another issue? My changes here aren't really related.

thoven87: I think this thread can be closed too as the ticket was resolved.

Sources/Jobs/AnyDecodableJob.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+/// Add codable support for decoding any JobInstance
+internal struct AnyDecodableJob: DecodableWithUserInfoConfiguration, Sendable {
+    typealias DecodingConfiguration = JobRegistry
+
+    init(from decoder: Decoder, configuration register: DecodingConfiguration) throws {
+        let container = try decoder.container(keyedBy: _JobCodingKey.self)
+        let key = container.allKeys.first!
```
Joannis, line 21: Can you explain the superdecoder dance you're doing here? Is this supposed to decode siblings?

Sources/Jobs/EncodableJob.swift
Joannis, line 18:
```suggestion
struct EncodableJob<Parameters: Codable & Sendable>: Encodable, Sendable {
```

Sources/Jobs/JobQueue.swift
```diff
@@ -45,7 +45,7 @@ public struct JobQueue<Queue: JobQueueDriver>: Service {
     @discardableResult public func push<Parameters: Codable & Sendable>(
         id: JobIdentifier<Parameters>, parameters: Parameters
     ) async throws -> Queue.JobID {
-        let jobRequest = JobRequest(id: id, parameters: parameters)
+        let jobRequest = EncodabeJob(id: id, parameters: parameters, queuedAt: .now)
```
Joannis, line 48:
```suggestion
        let jobRequest = EncodableJob(id: id, parameters: parameters, queuedAt: .now)
```

Sources/Jobs/AnyDecodableJob.swift
Joannis: Is it possible to get rid of the force-unwrap in trade for throwing an error?

Joannis (changes requested):
Looks good so far, just some clarification and some typos

Sources/Jobs/AnyDecodableJob.swift
adam-fowler: Added comment and got rid of force unwrap

Joannis (approved):

Issue #12, opened 2024-08-21
[Features] Improvements to the job service


thoven87:
I think it would be nice to have to the following features:

1 - Retry back-off, currently a job will retry right after failure until it reaches the `maxRetryCount: SOME_VALUE`, it would be great to introduce a exponential back off strategy.

2 - Delay until, the current implementation process a job right after it was queued. Most of the time it's what we want, but sometimes we want to process a task later. I think adding a delay until can be beneficial with suggestion number 1.

3 - Scheduled Jobs. It would be nice [PR](#5) landed.

I suppose with the current implementation, one needs Just one Job queue controller to register all jobs. Moreover, If we want to process jobs on another server, we would just not register the job queue service in the service lifecycle?

adam-fowler:
> I think it would be nice to have to the following features:
> 
> 1 - Retry back-off, currently a job will retry right after failure until it reaches the `maxRetryCount: SOME_VALUE`, it would be great to introduce a exponential back off strategy.
> 

Yes this is something we should do.

> 2 - Delay until, the current implementation process a job right after it was queued. Most of the time it's what we want, but sometimes we want to process a task later. I think adding a delay until can be beneficial with suggestion number 1.
> 

This is more complex, and at least for Redis will possibly require some work. But should be doable.

> 3 - Scheduled Jobs. It would be nice [PR](#5) landed.
> 
> I suppose with the current implementation, one needs Just one Job queue controller to register all jobs. Moreover, If we want to process jobs on another server, we would just not register the job queue service in the service lifecycle?

With the current implementation the scheduler is meant to run on one server and it will add jobs to the queue at scheduled times. Then one of the machines that is servicing the job queue will run the job.

thoven87:
> > I think it would be nice to have to the following features:
> > 1 - Retry back-off, currently a job will retry right after failure until it reaches the `maxRetryCount: SOME_VALUE`, it would be great to introduce a exponential back off strategy.
> 
> Yes this is something we should do.
> Sweet, I am experimenting with exponential back off for retry in a branch.
> 
> > 2 - Delay until, the current implementation process a job right after it was queued. Most of the time it's what we want, but sometimes we want to process a task later. I think adding a delay until can be beneficial with suggestion number 1.
> 
> This is more complex, and at least for Redis will possibly require some work. But should be doable.
> I don't know what the impact will be on the Redis job queue yet, I think adding delayUtil in the QueueJob struct can be reused for both schedule later and back off retry.
[...]

PR #21, opened 2024-08-30
feat(retry with backoff): Retry with backoff

thoven87:
First pass at adding retry with exponential backoff

I don't like the this implementation for the following reasons:

1 - The current execution will sleep
2 - A Job might cause the current execution to sleep for a while based on the maxAttempt passed.

I think we should update the Encodable job structure with the followings

1 - Delayed until (Time when to process this job)
2 - Remaining attempt count 

Adding the above would also afford us with allowing a job to be queued now and get process later. 

I look forward to your feedback on how best to proceed here CC: @adam-fowler @Joannis

### DIFF
--- a/Sources/Jobs/EncodableJob.swift
+++ b/Sources/Jobs/EncodableJob.swift
@@ -19,18 +19,9 @@ struct EncodableJob<Parameters: Codable & Sendable>: Encodable, Sendable {
     let id: JobIdentifier<Parameters>
     let data: JobInstanceData<Parameters>
 
-    init(
-        id: JobIdentifier<Parameters>,
-        parameters: Parameters,
-        queuedAt: Date,
-        attempts: Int
-    ) {
+    init(id: JobIdentifier<Parameters>, parameters: Parameters, queuedAt: Date, attempts: Int) {
         self.id = id
-        self.data = .init(
-            parameters: parameters,
-            queuedAt: queuedAt,
-            attempts: attempts
-        )
+        self.data = .init(parameters: parameters, queuedAt: queuedAt, attempts: attempts)
     }
 
     func encode(to encoder: Encoder) throws {
--- a/Sources/Jobs/EncodableJob.swift
+++ b/Sources/Jobs/EncodableJob.swift
@@ -19,9 +19,20 @@ struct EncodableJob<Parameters: Codable & Sendable>: Encodable, Sendable {
     let id: JobIdentifier<Parameters>
     let data: JobInstanceData<Parameters>
 
-    init(id: JobIdentifier<Parameters>, parameters: Parameters, queuedAt: Date) {
+    init(
+        id: JobIdentifier<Parameters>,
+        parameters: Parameters,
+        queuedAt: Date,
+        delayUntil: Date?,
+        attempts: Int
+    ) {
         self.id = id
-        self.data = .init(parameters: parameters, queuedAt: queuedAt)
+        self.data = .init(
+            parameters: parameters,
+            queuedAt: queuedAt,
+            delayUntil: delayUntil,
+            attempts: attempts
+        )
     }
 
     func encode(to encoder: Encoder) throws {
--- a/Sources/Jobs/JobInstance.swift
+++ b/Sources/Jobs/JobInstance.swift
@@ -24,6 +24,10 @@ protocol JobInstanceProtocol: Sendable {
     var maxRetryCount: Int { get }
     /// Time job was queued
     var queuedAt: Date { get }
+    /// Time to execute a job
+    var delayUntil: Date? { get }
+    /// Number of attempts so far
+    var attempts: Int { get }
     /// Function to execute the job
     func execute(context: JobContext) async throws
 }
@@ -32,6 +36,10 @@ extension JobInstanceProtocol {
     /// name of job type
     public var name: String {
         id.name
+    }
+
+    public var remainingAttempts: Int {
+        maxRetryCount - attempts
     }
 }
 
@@ -44,13 +52,16 @@ struct JobInstance<Parameters: Codable & Sendable>: JobInstanceProtocol {
     let job: JobDefinition<Parameters>
     /// job parameters
     let data: JobInstanceData<Parameters>
-
     /// job identifier
     var id: JobIdentifier<Parameters> { self.job.id }
     /// max retry count for a job
     var maxRetryCount: Int { self.job.maxRetryCount }
     /// Time job was queued
     var queuedAt: Date { self.data.queuedAt }
+    /// Number of attempts so far
+    var attempts: Int { self.data.attempts }
+    /// When to execute a job
+    var delayUntil: Date? { self.data.delayUntil }
 
     func execute(context: JobContext) async throws {
         try await self.job.execute(self.data.parameters, context: context)
@@ -68,10 +79,16 @@ struct JobInstanceData<Parameters: Codable & Sendable>: Codable {
     let parameters: Parameters
     /// Date job was queued
     let queuedAt: Date
+    /// When to execute a job in the future
+    let delayUntil: Date?
+    /// Number of attempts so far
+    let attempts: Int
 
     // keep JSON strings small to improve decode speed
     private enum CodingKeys: String, CodingKey {
         case parameters = "p"
         case queuedAt = "q"
+        case delayUntil = "d"
+        case attempts = "a"
     }
 }
--- a/Sources/Jobs/JobInstance.swift
+++ b/Sources/Jobs/JobInstance.swift
@@ -25,7 +25,7 @@ protocol JobInstanceProtocol: Sendable {
     /// Time job was queued
     var queuedAt: Date { get }
     /// Number of attempts so far
-    var attempts: Int { get }
+    var attempts: Int? { get }
     /// Job parameters
     var parameters: Parameters { get }
     /// Function to execute the job
@@ -40,12 +40,12 @@ extension JobInstanceProtocol {
 
     /// Number of remaining attempts
     public var remainingAttempts: Int {
-        maxRetryCount - attempts
+        maxRetryCount - (attempts ?? 0)
     }
 
     /// If job failed after n number of attempts
     public var didFail: Bool {
-        attempts >= maxRetryCount
+        (attempts ?? 0) >= maxRetryCount
     }
 }
 
@@ -65,7 +65,7 @@ struct JobInstance<Parameters: Codable & Sendable>: JobInstanceProtocol {
     /// Time job was queued
     var queuedAt: Date { self.data.queuedAt }
     /// Number of attempts so far
-    var attempts: Int { self.data.attempts }
+    var attempts: Int? { self.data.attempts ?? 0 }
     /// Job parameters
     var parameters: Parameters { self.data.parameters }
 
@@ -86,7 +86,7 @@ struct JobInstanceData<Parameters: Codable & Sendable>: Codable {
     /// Date job was queued
     let queuedAt: Date
     /// Number of attempts so far
-    let attempts: Int
+    let attempts: Int?
 
     // keep JSON strings small to improve decode speed
     private enum CodingKeys: String, CodingKey {
--- a/Sources/Jobs/JobInstance.swift
+++ b/Sources/Jobs/JobInstance.swift
@@ -24,22 +24,28 @@ protocol JobInstanceProtocol: Sendable {
     var maxRetryCount: Int { get }
     /// Time job was queued
     var queuedAt: Date { get }
-    /// Time to execute a job
-    var delayUntil: Date? { get }
     /// Number of attempts so far
     var attempts: Int { get }
+    /// Job parameters
+    var parameters: Parameters { get }
     /// Function to execute the job
     func execute(context: JobContext) async throws
 }
 
 extension JobInstanceProtocol {
-    /// name of job type
+    /// Name of job type
     public var name: String {
         id.name
     }
 
+    /// Number of remaining attempts
     public var remainingAttempts: Int {
         maxRetryCount - attempts
+    }
+
+    /// If job failed after n number of attempts
+    public var didFail: Bool {
+        attempts >= maxRetryCount
     }
 }
 
@@ -60,8 +66,8 @@ struct JobInstance<Parameters: Codable & Sendable>: JobInstanceProtocol {
     var queuedAt: Date { self.data.queuedAt }
     /// Number of attempts so far
     var attempts: Int { self.data.attempts }
-    /// When to execute a job
-    var delayUntil: Date? { self.data.delayUntil }
+    /// Job parameters
+    var parameters: Parameters { self.data.parameters }
 
     func execute(context: JobContext) async throws {
         try await self.job.execute(self.data.parameters, context: context)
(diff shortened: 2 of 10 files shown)


thoven87:
I opened a [ticket](https://github.com/hummingbird-project/swift-jobs/issues/12) to gathered some thoughts. Item number 3 can be marked as completed from the [JobScheduler PR](https://github.com/hummingbird-project/swift-jobs/pull/5). Adding delay until and remaining count would close item number 2.

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -121,6 +121,9 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
                     count -= 1
                     logger.debug("Retrying Job")
                     self.updateJobMetrics(for: job.name, startTime: startTime, retrying: true)
+                    /// Should we enque the job with delay until instead of sleeping?
+                    /// This would give us the benefit of allowing jobs to be pushed now and get processed later
+                    try await Task.sleep(nanoseconds: self.backoff(attempts: count))
```
thoven87, line 126: Instead of sleeping, I think the followings should be done instead

```swift
self.queue.push(
    Job(
        id,
        parameters,
        delayUntil: Date(timeIntervalSince1970: self.backoff(attempts: count)),
        remainingAttempt: count
    )
)
```

Then after the decoding step above, we should guard

```swift
guard (delayUntil ?? .distantPast) < Date() else {
    logger.info("requeing job because it's not ready to process.")
    self.queue.push(
        Job(
            id,
            parameters,
            delayUntil: Date(timeIntervalSinceNow: self.backoff(attempts: count)),
            remainingAttempt: count
        )
    )
}
```

adam-fowler: Yeah if we do a retry with back off then yeah pushing the jobs back onto the queue would be preferable. But I haven't thought of a nice way to do delayed jobs in redis yet.

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -112,15 +121,33 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
...
                     self.updateJobMetrics(for: job.name, startTime: startTime, retrying: true)
+                    let attempts = job.attempts + 1
+                    let delay = self.backoff(attempts: attempts)
+                    // let nextJobQueueDuration = Date.now.timeIntervalSince(job.queuedAt) + delay
+                    // once this is parameters are figured out. extact into a function so it can be used for both
+                    // delayed and retries
+                    let jobBuffer = EncodableJob(
+                        id: JobIdentifier(job.name),
+                        // TODO: How do I get job parameters from here?
+                        parameters: queuedJob,
```
thoven87, line 141: @adam-fowler or @Joannis I need some help here  please when you have a chance. How can I get access to job parameters here? 

I have want to move this operation into a function so that it can be used [here too](https://github.com/hummingbird-project/swift-jobs/pull/21/files#diff-ca3bd900c8cd6d93c3f9a8a1794fc50dcd394edd6351f2bf9acfb42d564ec3e7R106)

adam-fowler: Add to `JobInstanceProtocol`
```swift
    /// Job parameters
    var parameters: Parameters { get }
```
and add to `JobInstance`
```swift
    /// job parameters
    var parameters: Parameters { self.data.parameters }
```
then they are accessible as `job.parameters`

adam-fowler:
I would like to see a separate PR that details how delayed jobs are going to work.

I had a look to see how Vapor did it and I'm not overly keen on how it is done there. Basically a job is popped off the queue, and if it isn't ready to be processed then it is pushed back onto the queue. If the job queue only has jobs that are delayed it means every tick  (1 ms by default I think) a job will be popped off the queue and then pushed back. This seems like unnecessary work. But unless we can come up with an alternative I guess that is how we'll have to proceed.

thoven87:
> I would like to see a separate PR that details how delayed jobs are going to work.
> 
I am trying to figure out how to break this PR into two. I also wanted to implement the retry back off with the same logic that a delayed job would use. 
> I had a look to see how Vapor did it and I'm not overly keen on how it is done there. Basically a job is popped off the queue, and if it isn't ready to be processed then it is pushed back onto the queue. If the job queue only has jobs that are delayed it means every tick (1 ms by default I think) a job will be popped off the queue and then pushed back. This seems like unnecessary work. But unless we can come up with an alternative I guess that is how we'll have to proceed.

The initial idea I had was as follows:

## Postgres Job
1 - Either extend the current job table to add a new column called scheduled_at 
And then the query would change to add 

```SQL
WHERE schedued_at <= NOW()  +  INTERVAL ' 5 minutes' -- setting 5 minutes grace period here, we might need longer incase a job has not run in a while
```
2 - Add yet another table specifically for delayed jobs

## Redis
I am not too familiar with the Redis Stack library, but I think the following should work

keys
key,                             value,    TTL
swift_jobs::ready -> data -> -1
swift_jobs::queued->data->delayed_until_value -- Here we can use a Redis script that can atomically move jobs form queued to ready. The the script should be opaque to most since it's not swift code.
we could even add for job retry, but the above should work for retry with backoff
swift_jobs_::retry::count ->data->-1

Sources/Jobs/JobInstance.swift
```diff
@@ -68,10 +79,16 @@ struct JobInstanceData<Parameters: Codable & Sendable>: Codable {
     let parameters: Parameters
     /// Date job was queued
     let queuedAt: Date
+    /// When to execute a job in the future
+    let delayUntil: Date?
+    /// Number of attempts so far
+    let attempts: Int
```
Joannis, line 85: We need to look into this, because the non-optional field `attempts` will definitely be a breaking change.

Sources/Jobs/EncodableJob.swift
```diff
@@ -19,9 +19,9 @@ struct EncodableJob<Parameters: Codable & Sendable>: Encodable, Sendable {
     let id: JobIdentifier<Parameters>
     let data: JobInstanceData<Parameters>
 
-    init(id: JobIdentifier<Parameters>, parameters: Parameters, queuedAt: Date) {
+    init(id: JobIdentifier<Parameters>, parameters: Parameters, queuedAt: Date, attempts: Int) {
```
thoven87, line 22: should attempts get a default value to prevent jobs without this parameter from failing to decode?

adam-fowler (commented):
It's a good start. I'm not certain we need the addition protocol requirement for retrying a job though.

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -112,20 +114,36 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
...
                     }
-                    count -= 1
-                    logger.debug("Retrying Job")
-                    self.updateJobMetrics(for: job.name, startTime: startTime, retrying: true)
+
+                    let attempts = job.attempts + 1
+
+                    let delay = self.backoff(attempts: attempts)
+
+                    try await self.queue.retry(
```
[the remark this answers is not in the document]
thoven87: I thought of the retry as updating the current state of a job instead of appending a job to the queue. The Postgres implementation would update the job buffer. Now, that I think of it it could insert and oncmlict do update instead.

Sources/Jobs/JobQueueDriver.swift
```diff
@@ -45,4 +50,22 @@ public protocol JobQueueDriver: AsyncSequence, Sendable where Element == QueuedJ
...
+        )
+        return try JSONEncoder().encodeAsByteBuffer(data, allocator: ByteBufferAllocator())
+    }
+
+    func encode<Parameters: Codable & Sendable>(
+        id: JobIdentifier<Parameters>,
+        parameters: Parameters
+    ) throws -> (buffer: ByteBuffer, jobName: String) {
+        let jobRequest = EncodableJob(id: id, parameters: parameters, queuedAt: .now, attempts: 0)
+        return try (buffer: JSONEncoder().encodeAsByteBuffer(jobRequest, allocator: ByteBufferAllocator()), jobName: id.name)
```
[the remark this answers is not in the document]
thoven87: Will address in the next commit.

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -196,3 +214,28 @@ extension JobQueueHandler: CustomStringConvertible {
...
+        Double.random(in: 5.0..<15.0)
+    }
+
+    func backoff(
+        attempts: Int,
+        initialBackoff: TimeInterval = 1.0,
+        maximumBackoff: TimeInterval = 120.0
+    ) -> Date {
+        // Is this the best way to detect that we are in test mode?
+        let jitter = if NSClassFromString("XCTest") != nil {
```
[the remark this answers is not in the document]
thoven87: I will introduce yet another options structure then. Say, `JobHandlerOptions`?

Sources/Jobs/EncodableJob.swift
thoven87: I suppose we do need to pass `attempts` since data is of type `JobInstanceData` which defines `attempts`?

Sources/Jobs/MemoryJobQueue.swift
```diff
@@ -24,11 +24,13 @@ public final class MemoryQueue: JobQueueDriver {
     /// queue of jobs
     fileprivate let queue: Internal
     private let onFailedJob: @Sendable (QueuedJob<JobID>, any Error) -> Void
+    private let onRetryJob: @Sendable (QueuedJob<JobID>) -> Void
```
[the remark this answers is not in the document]
thoven87: remnants from proof of concept.

Sources/Jobs/JobQueueHandler.swift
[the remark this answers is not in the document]
thoven87: I thought of this a bit more. If the retry func isn't welcome, the push function will need updated to take in a job ID as most persisted drivers will need this id to update the delayed until parameters and buffer. Moreover, if we don't want to embed `attempts` in the job buffer, we would need a solution for the Redis driver. The Postgres driver would just need one more column.

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -112,20 +115,36 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
                     self.updateJobMetrics(for: job.name, startTime: startTime, error: error)
                     return
                 } catch {
-                    if count <= 0 {
-                        logger.debug("Job failed")
+                    if job.didFail {
+                        logger.debug("Job: failed")
```
Joannis, line 119: Idea: Can we maybe add the job ID and/or typename to the logs?

Joannis (approved):

Sources/Jobs/JobQueueHandler.swift
thoven87: JobId is already attached in the logger metadata [here](https://github.com/hummingbird-project/swift-jobs/blob/main/Sources/Jobs/JobQueueHandler.swift#L75)

Sources/Jobs/JobQueueHandler.swift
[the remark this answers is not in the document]
adam-fowler: The way I saw it was we are creating a new job. eg
```swift
self.queue.finish(jobId)
self.queue.push(jobData, options: .init(delayUntil: delay))
```
Replacing retry() with the above means we still pass all the tests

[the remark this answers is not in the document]
adam-fowler: We may need an `updateJob` function at some point though. But I don't think we need it here.

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -94,7 +98,6 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
             preferredDisplayUnit: .seconds
         ).recordSeconds(jobQueuedDuration)
 
-        var count = job.maxRetryCount
         logger.debug("Starting Job")
 
         do {
```
adam-fowler, line 103: We no longer need the while loop here anymore.

Sources/Jobs/JobQueueHandler.swift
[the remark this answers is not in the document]
thoven87: I see! For some reason, I didn't see this message until now.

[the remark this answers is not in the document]
thoven87: This has been addressed!

Sources/Jobs/JobQueueOptions.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import struct Foundation.TimeInterval
+
+/// JobQueueOptions
+public struct JobQueueOptions: Sendable {
+    /// Backoff factor default 2, but isn't used
+    /// since we ar esuing exp2 instead of pow
+    public var backoffFactor: Double
```
adam-fowler, line 21: If this isn't used should we include it

Sources/Jobs/JobQueueOptions.swift
```diff
@@ -0,0 +1,44 @@
...
+    /// Maximum jitter - default is 15 seconds
+    public var maxJitter: TimeInterval
+    /// Minimum jitter - default is 5 seconds
+    public var minJitter: TimeInterval
+    /// Computes a random jitter between minJitter and maxJitter
+    internal var jitter: Double {
+        Double.random(in: self.minJitter..<self.maxJitter)
+    }
+
+    public init(
```
adam-fowler, line 33: Can we add a documentation comment

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -94,34 +96,44 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
...
+                try await self.queue.finished(jobId: queuedJob.id)
+                // push new job in the queue
+                _ = try await self.queue.push(
+                    self.queue.encode(job, attempts: attempts),
+                    options: .init(
+                        delayUntil: delay
+                    )
+                )
+                self.updateJobMetrics(for: job.name, startTime: startTime, retrying: true)
+                logger.debug("Retrying Job with attempts: \(attempts) and delayed until: \(delay)")
```
adam-fowler, line 135: Logger messages should really be static and any dynamic elements be metadata eg
```suggestion
                logger.debug(
                    "Retrying Job", 
                    metadata: [
                        "attempts": .stringConvertible(attempts), 
                        "delayedUntil": .stringConvertible(delay)
                    ]
                )
```

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -196,3 +209,11 @@ extension JobQueueHandler: CustomStringConvertible {
         ).increment()
     }
 }
+
+extension JobQueueHandler {
+    func backoff(attempts: Int) -> Date {
+        let exp = Double(exp2(Double(attempts)))
```
adam-fowler, line 215: Does `exp2` not already return a Double

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -196,3 +209,11 @@ extension JobQueueHandler: CustomStringConvertible {
         ).increment()
     }
 }
+
+extension JobQueueHandler {
+    func backoff(attempts: Int) -> Date {
```
adam-fowler, line 214:
```suggestion
    func calculateBackoff(attempts: Int) -> Date {
```

adam-fowler (changes requested):
This looks good, just some minor cleanup stuff left

adam-fowler (approved):
Looks good

PR #23, opened 2024-09-04
feat(Delayed Job): Adding delayed job

thoven87:
This PR adds delayed Job. I will opened PR for the Postgres Job driver, then Redis next.


Sources/Jobs/MemoryJobQueue.swift
```diff
@@ -72,6 +72,7 @@ public final class MemoryQueue: JobQueueDriver {
     fileprivate actor Internal {
         var queue: Deque<QueuedJob<JobID>>
         var pendingJobs: [JobID: ByteBuffer]
+        var delayedJobs: [JobID: Date]
```
adam-fowler, line 75: Would it not be cleaner to include the delay date in the queue instead of a separate delayedJobs dictionary. 
```
var queue: Deque<(job: QueuedJob<JobID>, delayedUntil: Date?)>
```
This would mean when you pop a job off the queue that is not ready you are only editing the queue when you place it back on the queue.

Sources/Jobs/MemoryJobQueue.swift
```diff
@@ -104,6 +109,13 @@ public final class MemoryQueue: JobQueueDriver {
                     return nil
                 }
                 if let request = queue.popFirst() {
+                    if let delayedJob = self.delayedJobs.removeValue(forKey: request.id) {
+                        if delayedJob > Date() {
```
adam-fowler, line 113: Can we use Date.now is clearer

Tests/JobsTests/JobsTests.swift
```diff
@@ -137,6 +137,19 @@ final class JobsTests: XCTestCase {
...
+        let jobIdentifer = JobIdentifier<Int>(#function)
+        let expectation = XCTestExpectation(description: "TestJob.execute was called")
+        let jobQueue = JobQueue(.memory, numWorkers: 1, logger: Logger(label: "JobsTests"))
+        jobQueue.registerJob(id: jobIdentifer) { _, _ in
+            expectation.fulfill()
+        }
+        try await self.testJobQueue(jobQueue) {
+            try await jobQueue.push(id: jobIdentifer, parameters: 0, delayedUntil: .now.addingTimeInterval(5))
+            await fulfillment(of: [expectation], timeout: 10)
+        }
```
adam-fowler, line 150: This isn't really testing the job was delayed. It only tests it runs. Perhaps you should run two jobs, the first one is delayed and the second one isn't. Then make sure the second one ran first.

adam-fowler (changes requested):

Sources/Jobs/MemoryJobQueue.swift
Joannis, line 113: Bit of a nit, but I think it should be `>=`

```suggestion
                        if delayedJob >= Date() {
```

Sources/Jobs/JobQueue.swift
```diff
@@ -43,12 +43,14 @@ public struct JobQueue<Queue: JobQueueDriver>: Service {
     ///   - parameters: parameters for the job
     /// - Returns: Identifier of queued job
     @discardableResult public func push<Parameters: Codable & Sendable>(
-        id: JobIdentifier<Parameters>, parameters: Parameters
+        id: JobIdentifier<Parameters>,
+        parameters: Parameters,
+        delayedUntil: Date? = nil
```
Joannis, line 48: I'm happy to have more features, but I don't think every feature the job queue gets should be tacked onto a function signature like this. That will quickly explode into 5 or 10 more parameters if we let it

Joannis (changes requested):
We should look for a design pattern that allows expanding the job queue's feature set.
Possibly something like this:

```swift
public struct JobSchedulingOptions: ExpressibleByArrayLiteral {
  public struct Option {
    internal enum _Option {
      case delayed(until: Date)
    }
    
    internal let raw: _Option
    
    public static func delay(until date: Date) -> Option {
      Option(raw: .delayed(until: date))
    }
  }
  
  internal let _options: [Option]

  public init(arrayLiteral options: Option...) {
     self._options = options
  }
}
```

Sources/Jobs/MemoryJobQueue.swift
thoven87: That's much cleaner indeed!

Sources/Jobs/MemoryJobQueue.swift
thoven87: Yes, will do.

Tests/JobsTests/JobsTests.swift
thoven87: I will add another job without any delayed and check the execution order.

thoven87:
> We should look for a design pattern that allows expanding the job queue's feature set. Possibly something like this:
> 
> ```swift
> public struct JobSchedulingOptions: ExpressibleByArrayLiteral {
>   public struct Option {
>     internal enum _Option {
>       case delayed(until: Date)
>     }
>     
>     internal let raw: _Option
>     
>     public static func delay(until date: Date) -> Option {
>       Option(raw: .delayed(until: date))
>     }
>   }
>   
>   internal let _options: [Option]
> 
>   public init(arrayLiteral options: Option...) {
>      self._options = options
>   }
> }
> ```

@Joannis I would love to learn more about this design pattern you're suggesting here. I suppose, the push function would take a scheduler option and the delayed job would get executed on a scheduled job? The idea I had here was the followings:

Supposed you wanted to delay a job's execution for which ever reason (Welcome email e.t.c), would delay a job for 5 minutes, 1 hour or 3days which ever suits the reason for the delay. 

The next PR which is retry with backoff can reuse this same delay execution option. 

Yes, we currently, have options to have a scheduled job that runs every minute, hour and so forth. We could just spin off a schedule job to do the exact same thing at a given schedule. We would still need a delay option to retry jobs, specially jobs that make external API requests that might be rate limit. 

I would love to understand the proposal above further. Thanks!

Tests/JobsTests/JobsTests.swift
thoven87: The test has been updated to confirm that a delayed job is executed after a non delayed job.

Sources/Jobs/JobQueue.swift
```diff
@@ -43,12 +43,14 @@ public struct JobQueue<Queue: JobQueueDriver>: Service {
     ///   - parameters: parameters for the job
     /// - Returns: Identifier of queued job
     @discardableResult public func push<Parameters: Codable & Sendable>(
-        id: JobIdentifier<Parameters>, parameters: Parameters
+        id: JobIdentifier<Parameters>,
+        parameters: Parameters,
+        executionOptions: JobExecutionOptions = [.unknown()]
```
adam-fowler, line 48: Can we just call this `options`

Sources/Jobs/JobExecutionOptions.swift
```diff
@@ -0,0 +1,85 @@
...
+    /// Options  a job was queued with
+    public func options() -> [Option] {
+        return self._options
+    }
+
+    /// Time for when a job should run
+    public func delayedUntil() -> Date? {
+        return self._options.first(where: {
+            $0.isDelayed()
+        })?.delayedUntil()
```
adam-fowler, line 83: The fact we have to check twice if an option is a delayedUntil (in isDelayed, delayedUntil) can be avoided.
```swift
for option in options {
    switch option.raw {
        if case .delayed(let date) = option {
            return date
        }
    }
    return nil
}
```

Sources/Jobs/JobExecutionOptions.swift
```diff
@@ -0,0 +1,85 @@
...
+
+import struct Foundation.Date
+
+/// Job execution type
+public struct JobExecutionOptions: ExpressibleByArrayLiteral, Sendable {
+    /// Job execution options
+    public struct Option: Sendable {
+        internal enum _Option {
+            case delayed(until: Date)
+            case unknown
```
adam-fowler, line 23: What is unknown for?

Tests/JobsTests/JobsTests.swift
```diff
@@ -137,6 +137,36 @@ final class JobsTests: XCTestCase {
...
+            jobExecutionSequence.withLockedValue {
+                $0.append(parameters)
+            }
+            expectation.fulfill()
+            delayedJob.wrappingDecrement(by: 1, ordering: .relaxed)
+        }
+        try await self.testJobQueue(jobQueue) {
+            delayedJob.wrappingIncrement(by: 1, ordering: .relaxed)
+            try await jobQueue.push(id: job1, parameters: 0, executionOptions: [
+                .delay(until: Date.now.addingTimeInterval(5)),
```
adam-fowler, line 158: 5 seconds is long for a test, can we reduce this to 1.

adam-fowler (changes requested):
Looks good, a few minor nits

Sources/Jobs/JobExecutionOptions.swift
thoven87: > The fact we have to check twice if an option is a delayedUntil (in isDelayed, delayedUntil) can be avoided.
> 
> ```swift
> for option in options {
>     switch option.raw {
>         if case .delayed(let date) = option {
>             return date
>         }
>     }
>     return nil
> }
> ```

I think I can get rid of the helper functions I had defined before then: `isDelayed()`, `isUnknown()` and `delayedUntil()`

Sources/Jobs/JobExecutionOptions.swift
thoven87: I think it's remnants when I was learning about `ExpressibleByArrayLiteral`. It's not needed since an empty array can set as default.

Sources/Jobs/MemoryJobQueue.swift
```diff
@@ -46,8 +46,8 @@ public final class MemoryQueue: JobQueueDriver {
     ///   - job: Job
     ///   - eventLoop: Eventloop to run process on (ignored in this case)
     /// - Returns: Queued job
-    @discardableResult public func push(_ buffer: ByteBuffer) async throws -> JobID {
-        return try await self.queue.push(buffer)
+    @discardableResult public func push(_ buffer: ByteBuffer, options: JobExecutionOptions) async throws -> JobID {
+        return try await self.queue.push(buffer, executionOptions: options)
```
adam-fowler, line 50: Can we propagate the `executionOptions` -> `options` change through all of the functions. Including the type name, should be `JobOptions`.

Joannis (approved):

Sources/Jobs/JobOptions.swift
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import struct Foundation.Date
+
+/// Job options
+public struct JobOptions: Sendable {
+    /// When to execute the job
+    public let delayUntil: Date?
```
Joannis, line 20: I'm of the camp of making things a `let` by default, even in structs, but think should make this a `var` instead

Joannis (approved):
Looks good to me

adam-fowler (approved):
Looks good, once we get the redis implementation sorted, I think we'll be able to do a mass merge of all of these PRs

thoven87:
> Looks good, once we get the redis implementation sorted, I think we'll be able to do a mass merge of all of these PRs

I opened the Redis driver implementation [PR](https://github.com/hummingbird-project/swift-jobs-redis/pull/5) last night.

Sources/Jobs/JobOptions.swift
thoven87: > I'm of the camp of making things a `let` by default, even in structs, but think should make this a `var` instead

Should I change it to a var instead? I was under the impression, a user would set `delayUntil` once when pushing a job and would not need to mutate it after.

PR #28, opened 2024-09-15
chore(bug): fixed a bug where processing step was over counting

thoven87:
This PR is for the followings:

1 -  fixed an issue where the queued meter would become negative during retries
~~2 - added number of workers to job metrics~~
     processing count can give the same metric
3 - added discarded jobs metrics

With the most recent commit:

Once can get prometheus UI or Grafana to update queue in realtime with the following query

```promql
count(swift_jobs_meter{status="queued"} unless on(jobID) (swift_jobs_meter{status="processing"})) or vector(0)
```

### DIFF
--- a/Sources/Jobs/JobQueue.swift
+++ b/Sources/Jobs/JobQueue.swift
@@ -50,7 +50,10 @@ public struct JobQueue<Queue: JobQueueDriver>: Service {
         let buffer = try self.queue.encode(id: id, parameters: parameters)
         let jobName = id.name
         let id = try await self.queue.push(buffer, options: options)
-        Meter(label: "swift.jobs.meter", dimensions: [("status", "queued"), ("name", jobName)]).increment()
+        Meter(label: JobMetricsHelper.meterLabel, dimensions: [
+            ("status", JobMetricsHelper.JobStatus.queued.rawValue),
+            ("name", jobName),
+        ]).increment()
         self.logger.debug(
             "Pushed Job",
             metadata: ["JobID": .stringConvertible(id), "JobName": .string(jobName)]
--- a/Sources/Jobs/JobQueue.swift
+++ b/Sources/Jobs/JobQueue.swift
@@ -52,7 +52,6 @@ public struct JobQueue<Queue: JobQueueDriver>: Service {
         let id = try await self.queue.push(buffer, options: options)
         Meter(label: JobMetricsHelper.meterLabel, dimensions: [
             ("status", JobMetricsHelper.JobStatus.queued.rawValue),
-            ("name", jobName),
         ]).increment()
         self.logger.debug(
             "Pushed Job",
--- a/Sources/Jobs/JobQueue.swift
+++ b/Sources/Jobs/JobQueue.swift
@@ -52,6 +52,7 @@ public struct JobQueue<Queue: JobQueueDriver>: Service {
         let id = try await self.queue.push(buffer, options: options)
         Meter(label: JobMetricsHelper.meterLabel, dimensions: [
             ("status", JobMetricsHelper.JobStatus.queued.rawValue),
+            ("jobID", id.description),
         ]).increment()
         self.logger.debug(
             "Pushed Job",
--- a/Sources/Jobs/JobQueue.swift
+++ b/Sources/Jobs/JobQueue.swift
@@ -48,9 +48,9 @@ public struct JobQueue<Queue: JobQueueDriver>: Service {
         options: JobOptions = .init()
     ) async throws -> Queue.JobID {
         let buffer = try self.queue.encode(id: id, parameters: parameters)
-        Meter(label: "swift.jobs.meter", dimensions: [("status", "queued")]).increment()
         let jobName = id.name
         let id = try await self.queue.push(buffer, options: options)
+        Meter(label: "swift.jobs.meter", dimensions: [("status", "queued"), ("name", jobName)]).increment()
         self.logger.debug(
             "Pushed Job",
             metadata: ["JobID": .stringConvertible(id), "JobName": .string(jobName)]
--- a/Sources/Jobs/JobQueueHandler.swift
+++ b/Sources/Jobs/JobQueueHandler.swift
@@ -67,18 +67,6 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
     func runJob(_ queuedJob: QueuedJob<Queue.JobID>) async throws {
         var logger = logger
         let startTime = DispatchTime.now().uptimeNanoseconds
-        // Decrement the current job by 1
-        Meter(label: JobMetricsHelper.meterLabel, dimensions: [
-            ("status", JobMetricsHelper.JobStatus.queued.rawValue),
-        ]).decrement()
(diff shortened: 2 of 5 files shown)


Sources/Jobs/JobQueueHandler.swift
```diff
@@ -181,6 +191,8 @@ extension JobQueueHandler: CustomStringConvertible {
                 label: self.metricsLabel,
                 dimensions: [("name", name), ("status", JobStatus.retried.rawValue)]
             ).increment()
+            // increment queued meter else we will get negative values
+            Meter(label: self.meterLabel, dimensions: [("status", "queued"), ("name", name)]).increment()
```
thoven87, line 195: cc @Joannis you were correct all along. `self.queue.push` is does not increment the meter 🥇

Sources/Jobs/JobQueue.swift
```diff
@@ -48,9 +48,9 @@ public struct JobQueue<Queue: JobQueueDriver>: Service {
         options: JobOptions = .init()
     ) async throws -> Queue.JobID {
         let buffer = try self.queue.encode(id: id, parameters: parameters)
-        Meter(label: "swift.jobs.meter", dimensions: [("status", "queued")]).increment()
         let jobName = id.name
         let id = try await self.queue.push(buffer, options: options)
+        Meter(label: "swift.jobs.meter", dimensions: [("status", "queued"), ("name", jobName)]).increment()
```
Joannis, line 53: I assume this was causing issues when a push fails?

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -27,6 +27,7 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
         self.logger = logger
         self.jobRegistry = .init()
         self.options = options
+        Meter(label: "swift.jobs.worker.count").increment(by: Double(numWorkers))
```
Joannis, line 30: When does this number decrement?

Joannis: What happens when an app crashes?

Sources/Jobs/JobQueue.swift
adam-fowler: The meter label is hard coded here, but not in JobQueueHandler

Sources/Jobs/JobQueueHandler.swift
adam-fowler: Ah that's because we call direct to the queue driver to push it back onto the queue

adam-fowler: Sometimes you use "queued" and sometimes you use `JobStatus.queued.rawValue`. Can we have some consistency

thoven87: > Ah that's because we call direct to the queue driver to push it back onto the queue

It's also possible to have negative values when the server boots. This case is only valid when the job was shutdown while there were still jobs to be processed in the queue. I think queued gauge inc should be moved to one of the push func that'll increment on init, and push to a queue driver. I think the former should only inc if there are jobs in the queue.

Sources/Jobs/JobQueue.swift
thoven87: I am moving the labels into an internal structure named `JobMetricsHelper` so they can be accessible in the swift jobs library.

Sources/Jobs/JobQueueHandler.swift
thoven87: We should not decrement this gauge. We can use it for alerting purposes because once an app crashes or shutdown, after a while prometheus will return 0

adam-fowler: > We should not decrement this gauge. We can use it for alerting purposes because once an app crashes or shutdown, after a while prometheus will return 0

Are you saying that prometheus will revert a Meter to zero if the connection from the app is lost due to a crash? What does that mean for all of the other Meters?

thoven87: I should have phrased the previous sentence better. Prometheus would not reset the meter to zero, there would be no data if one where to query for most recent update. 

```pql
max_over_time(swift_jobs_worker_count{}[5m]))
```

To be safe, I will try to decrement during graceful shutdown. I can see the worker count might be misleading in a continuous deployment environment.

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -68,54 +67,66 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
...
+            label: "\(JobMetricsHelper.metricsLabel).queued.duration",
             dimensions: [("name", job.name)],
             preferredDisplayUnit: .seconds
         ).recordSeconds(jobQueuedDuration)
 
+        // Decrement the current job by 1
+        Meter(label: JobMetricsHelper.meterLabel, dimensions: [
+            ("status", JobMetricsHelper.JobStatus.queued.rawValue),
+            ("name", job.name),
+        ]).decrement()
```
adam-fowler, line 105: I think this call should go at the top of the function, as the job has been removed from the queue by then and if the code fails to decode it doesnt decrement the metric.

adam-fowler (changes requested):
One minor change, but otherwise we are good.

Sources/Jobs/JobQueueHandler.swift
thoven87: I will move this call at the top. I still have one question about ways of pushing jobs in the queue. [There's this push func](https://github.com/hummingbird-project/swift-jobs/blob/main/Sources/Jobs/JobQueue.swift#L45) and job parameters [push](https://github.com/hummingbird-project/swift-jobs/blob/main/Sources/Jobs/JobParameters.swift#L28) which calls JobQueue [push](https://github.com/hummingbird-project/swift-jobs/blob/main/Sources/Jobs/JobParameters.swift#L39) I think we are missing queue inc between these funcs. JobScheduler [uses](https://github.com/hummingbird-project/swift-jobs/blob/main/Sources/Jobs/JobParameters.swift#L39). We might see negative count depending on how a job was pushed onto a queue.

adam-fowler: The user shouldn't be calling push on the driver, they need to do it through the queue. I'm not sure how we can be sure people don't do this though. In theory they could call any of the function on the driver and break stuff

thoven87: I see! Should the scheduler use this push [func](https://github.com/hummingbird-project/swift-jobs/blob/main/Sources/Jobs/JobQueue.swift#L45C36-L45C40) instead? Otherwise, there needs to be an Meter increment [here](https://github.com/hummingbird-project/swift-jobs/blob/main/Sources/Jobs/Scheduler/JobSchedule.swift#L187)

adam-fowler: The scheduler probably should but I think there was an issue getting that working. We might need to move the JobQueueDriver extension onto JobQueue if I remember right. Let's look into that in a separate PR. I'd really like to land this one and get a release out with the new swift-jobs-postgres repo as well.

adam-fowler (approved):

PR #31, opened 2024-09-17
feat(every): adding every N minutes or hours

thoven87:
This PR adds the followings:

Run a job every 5 minutes or 15, 30, 45 and so forth. Please note that all minutes or hours always start at 0

```swift
var jobScheduleService = JobSchedule()
jobScheduleService.addJob(
        ScheduledJobController.StatsParameters(count: 20),
        schedule: .every(minute: 5)
)
```

The same is true for every N hour

```swift
var jobScheduleService = JobSchedule()
jobScheduleService.addJob(
        ScheduledJobController.StatsParameters(count: 20),
        schedule: .every(hour: 3)
)
```

### DIFF
--- a/Sources/Jobs/Scheduler/Schedule.swift
+++ b/Sources/Jobs/Scheduler/Schedule.swift
@@ -117,6 +117,39 @@ public struct Schedule: Sendable {
         .init(second: .specific(second))
     }
 
+    /// Return schedule that generates a Date for minutes, hours or seconds
+    /// - Parameter minute: minute value e.g 5
+    /// - Parameter hour: hour value e.g 3
+    /// - Parameter timeZone defaults to GMT
+    public static func every(minute: Int = 5, hour: Int = 0, timeZone: TimeZone = .gmt) -> Self {
+        let maxNumber = if minute > 0, hour == 0 {
+            60 / minute
+        } else {
+            24 / hour
+        }
+
+        var multiples: [Int] = []
+        var index = 0
+        let number = hour > 0 ? hour : minute
+        repeat {
+            let multiple = number * index
+
+            if hour > 0, multiple > 24 {
+                let remainder = multiple % 24
+                multiples.append(remainder - hour)
+            } else {
+                multiples.append(multiple)
+            }
+
+            index = index + 1
+        } while index < maxNumber
+
+        if hour == 0 {
+            return .onMinutes(multiples)
+        }
+        return .onHours(multiples, timeZone: timeZone)
+    }
+
     /// Return Schedule that generates a Date for a selection of minutes
     /// - Parameters
     ///   - minutes: Array of minutes if should return Dates for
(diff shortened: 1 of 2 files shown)


adam-fowler (changes requested):
Two things
- This should be two functions, one for minutes and one for hours
- The function promises something it can't deliver ie generate an event every so many minutes. This only works for factors of 60 or 24 depending on value.

Perhaps you should have an enum for the factors eg
```swift
enum MinuteFactor: Int {
  case two = 2
  case three = 3
  case four = 4
  case five = 5
  case six = 6
  case ten = 10
  case twelve = 12
  case quarterHour = 15
  case twenty = 20
  case halfHour = 30
}
func everyMinute(_ minute: MinuteFactor)
```
And something similar for hours. For the hours you could also add a start time eg
```swift
func everyHour(_ hour: HourFactor, startingHour: Int, minute: Int)
```

thoven87:
> Two things
> 
> * This should be two functions, one for minutes and one for hours
> * The function promises something it can't deliver ie generate an event every so many minutes. This only works for factors of 60 or 24 depending on value.
> 
> Perhaps you should have an enum for the factors eg
> 
> ```swift
> enum MinuteFactor: Int {
>   case two = 2
>   case three = 3
>   case four = 4
>   case five = 5
>   case six = 6
>   case ten = 10
>   case twelve = 12
>   case quarterHour = 15
>   case twenty = 20
>   case halfHour = 30
> }
> func everyMinute(_ minute: MinuteFactor)
> ```
> 
> And something similar for hours. For the hours you could also add a start time eg
> 
> ```swift
> func everyHour(_ hour: HourFactor, startingHour: Int, minute: Int)
> ```

I need to re-think this PR a little bit. There's currently func called everyMinute which takes seconds, but does not work with fractions i.e everyMinute(second: 300) which would translate into every 5 minutes starting at 0-5-10-15-20...55

Would it be easier to expose a func that takes in a cron syntax and computes the schedule?

adam-fowler:
> I need to re-think this PR a little bit. There's currently func called everyMinute which takes seconds, but does not work with fractions i.e everyMinute(second: 300) which would translate into every 5 minutes starting at 0-5-10-15-20...55

The seconds parameter in `everyMinute` is when in the minute the schedule triggers a job. It is probably unnecessary for this function, but for `everyHour` or `everyDay` it is important to be able to say when in the hour or day we want the job to trigger. `everyMinute(second: 300)` basically means never as the second value can only be between 0 and 59.

> Would it be easier to expose a func that takes in a cron syntax and computes the schedule?

We don't support the full crontab syntax eg we don't support `* 4 * * *` ie every minute when the when the hour is 4

thoven87:
> > I need to re-think this PR a little bit. There's currently func called everyMinute which takes seconds, but does not work with fractions i.e everyMinute(second: 300) which would translate into every 5 minutes starting at 0-5-10-15-20...55
> 
> The seconds parameter in `everyMinute` is when in the minute the schedule triggers a job. It is probably unnecessary for this function, but for `everyHour` or `everyDay` it is important to be able to say when in the hour or day we want the job to trigger. `everyMinute(second: 300)` basically means never as the second value can only be between 0 and 59.
> 
> > Would it be easier to expose a func that takes in a cron syntax and computes the schedule?
> 
> We don't support the full crontab syntax eg we don't support `* 4 * * *` ie every minute when the when the hour is 4

Understood, I thought of writing a Cron parser and expose mundane syntax such as 

```swift
@daily
@weekly
@monthly
@yearly
@minute
```

and finally 
```sh
* 4 * * *
5 * * * *
* * * * * 
```
What do you think?

PR #46, opened 2025-01-02
Remove job id from metrics

adam-fowler:
It generates a unique metric for each job instance


thoven87 (approved):
Looks good! Should an issue be opened on how to extract job name and make it available in the meters? Or is this not necessary?

adam-fowler:
> Looks good! Should an issue be opened on how to extract job name and make it available in the meters? Or is this not necessary?

Add an issue.

PR #50, opened 2025-01-09
Metric cleanup

adam-fowler:
- Removed unnecessary meters
- Renamed metrics label variables

### DIFF
--- a/Sources/Jobs/JobMetricsHelper.swift
+++ b/Sources/Jobs/JobMetricsHelper.swift
@@ -16,12 +16,16 @@ import Metrics
 
 /// OTEL labels and dimensions
 internal enum JobMetricsHelper {
-    /// Metrics label
-    static let metricsLabel: String = "swift.jobs"
-    /// Meter label for Processsing, Queued, Failed and Completed
+    /// Counter label
+    static let counterLabel: String = "swift.jobs"
+    /// Job duration timer label
+    static let timerLabel: String = "swift.jobs.duration"
+    /// Job queued timer label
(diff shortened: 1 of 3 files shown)


Sources/Jobs/JobMetricsHelper.swift
```diff
@@ -46,21 +50,9 @@ internal enum JobMetricsHelper {
...
-        // Or decrement processing vector in Prometheus UI or Grafana
-        // with something like count(swif_jobs_meter{status="processing"}
-        // unless on(jobID) (swif_jobs_meter{status="queued"})
-        // or (swif_jobs_meter{status="completed")) or vector(0)
-        Meter(
-            label: JobMetricsHelper.meterLabel,
-            dimensions: [
-                ("status", JobMetricsHelper.JobStatus.completed.rawValue)
-            ]
-        ).increment()
```
adam-fowler, line 59: This value in this meter can be accessed from counter `swift.jobs` with dimension status set to completed

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -99,8 +99,8 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
         } catch let error as JobQueueError where error == .unrecognisedJobId {
             logger.debug("Failed to find Job with ID while decoding")
             try await self.queue.failed(jobId: queuedJob.id, error: error)
-            Meter(
-                label: JobMetricsHelper.discardedMeter,
+            Counter(
+                label: JobMetricsHelper.discardedCounter,
```
adam-fowler, line 103: Given the value never goes down we can convert this to a counter

PR #54, opened 2025-02-05
Job Middleware

adam-fowler:
See https://github.com/hummingbird-project/hummingbird-evolution/blob/jobs-middleware/proposals/NNNN-swift-jobs-hooks.md

I've added metrics and tracing middleware to show how this might work.


Sources/Jobs/Middleware/JobMiddleware.swift
```diff
@@ -0,0 +1,152 @@
...
+    ///
+    /// - Parameters:
+    ///   - job: Job instance
+    ///   - context: Job context
+    ///   - next: Next handler
+    /// - Throws:
+    func handleJob(job: any JobInstanceProtocol, context: JobContext, next: (any JobInstanceProtocol, JobContext) async throws -> Void) async throws
+}
+
+public struct NullJobMiddleware: JobMiddleware {
```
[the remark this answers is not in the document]
adam-fowler: The result builder returns a JobMiddleware. What should it return if we don't provide any.

thoven87:
Looks great so far. The only issue I see with the current state is that job name is not always available. This is very important when debugging as jobs (specially is a graphical user interface) a "named queued" concept could help here I think as we would always able to map how many failures a queue has had overtime, number of "dead jobs" (jobs which can't run for some reasons).

Otherwise, one can end up with a bunch of job id and no knowing what kind of jobs they are. I think the job middleware should be able to report the job name or the topic name for each event. Otherwise, the push event should return the job id which an external subscriber can keep track of dead letter topics and so forth. 

The Job middleware sounds like an event type of middleware. Are they any reasons why the "on" keyword is not used? I find the popJob handleJob naming a bit on confusing. should the handleJob func be renamed to something that describes that a job is running? reading handleJob makes me believe that the middleware will process the job.

adam-fowler:
> Looks great so far. The only issue I see with the current state is that job name is not always available.

The job name/id is available in every situation (either via JobIdentifier, or JobInstanceProtocol.id) except when pop job fails. If pop job fails there is no job, so there's no name.

> The Job middleware sounds like an event type of middleware. Are they any reasons why the "on" keyword is not used? I find the popJob handleJob naming a bit on confusing. should the handleJob func be renamed to something that describes that a job is running? reading handleJob makes me believe that the middleware will process the job.

Renamed `popJob` and `pushJob` to `onPopJob` and `onPushJob`.
`handleJob` is called `handleJob` because it does handle the processing of the job. You pass in the job.execute() call in a closure and it gets called at the end of the middleware chain.

thoven87:
> > Looks great so far. The only issue I see with the current state is that job name is not always available.
> 
> The job name/id is available in every situation (either via JobIdentifier, or JobInstanceProtocol.id) except when pop job fails. If pop job fails there is no job, so there's no name.
> 
> > The Job middleware sounds like an event type of middleware. Are they any reasons why the "on" keyword is not used? I find the popJob handleJob naming a bit on confusing. should the handleJob func be renamed to something that describes that a job is running? reading handleJob makes me believe that the middleware will process the job.
> 
> Renamed `popJob` and `pushJob` to `onPopJob` and `onPushJob`. `handleJob` is called `handleJob` because it does handle the processing of the job. You pass in the job.execute() call in a closure and it gets called at the end of the middleware chain.

Say we have the following 

```swift
struct ProviderETLParameters: Coddle {
  let partenerName: String
  ......
}

struct DataProviderETLJobParameters: JobParameters {
    static let jobName: String = "DataProviderETLHourlyJob"
    let parameters: ProviderETLParameters
}
```

Now, some job with the above parameters was enqueued and failed for some reason and it'll be retried after some time.

I decided to change the above parameters to 

```swift
struct ProviderETLParameters: Coddle {
  let partenerName: String
  ......
}

struct DataProvidersETLJob: JobParameters {
    static let jobName: String = "DataProvidersHourlyJob"
    let parameters: ProviderETLParameters
}
```

What would happen here since there will be no service to process this job would we still be able to decode and have the previous JobName? This would be an example of a "dead job"

if not, should we skip earlier if pop returns nil?

adam-fowler:
> What would happen here since there will be no service to process this job would we still be able to decode and have the previous JobName? This would be an example of a "dead job"

I guess you are talking about the situation where a decode fails. Of course you shouldn't be editing job parameters once they are out in the wild. At least you should only be allowed to add additional optional parameters. But yeah programmers make mistakes and need some support.

We could put an optional `jobName` in `JobQueueError` which would be filled out in the situations we had a name for example decode issues.

Sources/Jobs/Middleware/JobMiddleware.swift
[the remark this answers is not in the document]
Joannis: Hmm, in hindsight I think this is more performant than the solution I was thinking of - which was parameter-pack or array based

Sources/Jobs/Middleware/MetricsMiddleware.swift
```diff
@@ -0,0 +1,227 @@
...
+    static let timerLabel: String = "swift.jobs.duration"
+    /// Job queued timer label
+    @usableFromInline
+    static let queuedTimerLabel: String = "swift.jobs.queued.duration"
+    /// Meter label for Processsing, Queued
+    @usableFromInline
+    static let meterLabel: String = "swift.jobs.meter"
+    /// Counter label for discarded jobs
+    @usableFromInline
+    static let discardedCounter: String = "swift.jobs.discarded"
```
Joannis, line 35: Shouldn't it be labeled `jobs.swift` since it's swift-jobs? Just like it'd be `com.apple.my-service` rather than `apple.com.my-service`

Joannis:
Looks great!

Sources/Jobs/JobContext.swift
```diff
@@ -15,5 +15,8 @@
 import Logging
 
 public struct JobContext {
```
thoven87, line 17: Would it make sense to add job started, ended, elapse time here? the events middleware could propagate that downstream? Or would JobParameters be the place to put that information?

adam-fowler: Not sure why you would need to include these in the JobContext, given you can work them out. Also JobEnded is not available until the job has finished.

```swift
public struct MyJobMiddleware: JobMiddleware {
    public init() {}

    /// Handle job and pass it onto next handler (works like middleware in Hummingbird)
    @inlinable
    public func handleJob(
        job: any JobInstanceProtocol,
        context: JobContext,
        next: (any JobInstanceProtocol, JobContext) async throws -> Void
    ) async throws {
        let jobStarted = Date.now
        try await next(job, context)
        let jobEnded = Date.now
    }
}
```

thoven87: I just thought that patterns that are repeated would be nice to live "upstream".

Sources/Jobs/Middleware/MetricsMiddleware.swift
adam-fowler: These are the metrics we already had

Sources/Jobs/JobContext.swift
adam-fowler: We can discuss additional job context data, in a separate PR I think

Joannis (approved):

thoven87 (approved):

PR #57, opened 2025-02-13
Add tracing tests

adam-fowler:


Tests/JobsTests/TestTracer.swift
```diff
@@ -0,0 +1,182 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See hummingbird/CONTRIBUTORS.txt for the list of Hummingbird authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+//===----------------------------------------------------------------------===//
+//
```
thoven87, line 15: nit: it looks like the file header was generated twice?

thoven87 (commented):
I learned at least two new things reading this code `@autoclosure` and `private(set)`.

Tests/JobsTests/TestTracer.swift
adam-fowler: These are two different headers, one is from Swift Distributed Tracing and other Hummingbird

thoven87 (approved):

PR #60, opened 2025-02-18
Call queue.fail inside an unstructured Task

adam-fowler:
This is ensure Task cancellation isn't propagated to the fail call

### DIFF
--- a/Sources/Jobs/JobQueueHandler.swift
+++ b/Sources/Jobs/JobQueueHandler.swift
@@ -111,10 +111,13 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
                 }
             } catch let error as CancellationError {
                 logger.debug("Job cancelled")
-                // Job failed is called but due to the fact the task is cancelled, depending on the
-                // job queue driver, the process of failing the job might not occur because itself
-                // might get cancelled
-                try await self.queue.failed(jobId: queuedJob.id, error: error)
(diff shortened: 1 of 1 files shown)


Sources/Jobs/JobQueueHandler.swift
```diff
@@ -111,10 +111,13 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
...
-                // Job failed is called but due to the fact the task is cancelled, depending on the
-                // job queue driver, the process of failing the job might not occur because itself
-                // might get cancelled
-                try await self.queue.failed(jobId: queuedJob.id, error: error)
+                // We need to wrap the failed call in a unstructured Task to stop propagation of
+                // Task cancellation to the `queue.fail()` call. We are calling `get` on the Task
+                // as soon as we create it so can guarantee the Task is done when we leave the
+                // function.
+                try await Task {
+                    try await self.queue.failed(jobId: queuedJob.id, error: error)
```
thoven87, line 119: what happens if the call to move the jobs to failed sate fails? Should the cancellation reason be logged?

adam-fowler: If it doesn't get set to failed it will stay as processing. A CancellationError doesn't report a reason. All we know is the job queue service was cancelled at some point. ServiceLifecycle can do this if graceful shutdown takes too long.

thoven87 (approved):

PR #67, opened 2025-02-23
Adding queue name with default

thoven87:
* Adding queue name with default

I would some feedback on this As I think queue name should be available to the middleware too.


adam-fowler:
Why are we adding the queue name to the queue? The whole point of adding a queue name to the Postgres driver was so we can have multiple JobQueues with different setups (eg middleware, number of workers). You've now got one JobQueue with multiple queue names.

thoven87:
> Why are we adding the queue name to the queue? The whole point of adding a queue name to the Postgres driver was so we can have multiple JobQueues with different setups (eg middleware, number of workers). You've now got one JobQueue with multiple queue names.

Currently, there's no observability into queue names. I opened this PR because I wanted to expose metrics per queue perhaps, `JobQueueOptions` is the place to put `queueName` instead? we would still have access to the queue name via the jobQueue handler. Is this not needed?

PR #68, opened 2025-02-23
Job schedule fixes

adam-fowler:
Fix picking correct date 

Added a load of tests to ensure it works this time.

### DIFF
--- a/Sources/Jobs/Scheduler/JobSchedule.swift
+++ b/Sources/Jobs/Scheduler/JobSchedule.swift
@@ -99,12 +99,7 @@ public struct JobSchedule: MutableCollection, Sendable {
     }
 
     mutating func updateNextScheduledDate(jobIndex: Int) {
-        let dateFrom: Date =
-            switch self.self[jobIndex].accuracy {
-            case .latest: Swift.max(.now, self[jobIndex].nextScheduledDate)
-            case .all: self[jobIndex].nextScheduledDate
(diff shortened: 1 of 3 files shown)


Sources/Jobs/Scheduler/JobSchedule.swift
```diff
@@ -99,22 +97,41 @@ public struct JobSchedule: MutableCollection, Sendable {
...
         }
     }
 
-    mutating func setInitialNextDate(after date: Date) {
+    mutating func setInitialNextDate(after date: Date, now: Date = .now, logger: Logger) {
         for index in 0..<self.count {
-            self[index].nextScheduledDate = self[index].schedule.setInitialNextDate(after: date) ?? .distantFuture
+            switch self[index].accuracy {
+            case .all:
+                // set nextScheduledDate based on date supplied
```
thoven87, line 112: This is awesome, we now have backfill support built-in. So, in theory one can start a job with start date being `2025-01-01T00` with any scheduling scheme and the scheduler will enqueue all 'till now. I think this partition (schedule) should get propagated in the job context so that on will know which schedule that is currently running.

adam-fowler: Hard to do that given all the schedule does is push jobs onto the queue.

thoven87: I will explore what can be done here to make this happen

thoven87 (approved):

PR #69, opened 2025-02-24
Require all job parameters to conform to JobParameters

adam-fowler:
This PR also removes JobIdentifier as it is no longer needed

Also fix issue in MemoryJobQueue where if queue only existed of delayed jobs, it would get stuck in a loop until one of the jobs was no longer delayed.


Sources/Jobs/JobRegistry.swift
```diff
@@ -45,27 +45,17 @@ public final class JobRegistry: Sendable {
...
-    public func registerJob<Parameters: Codable & Sendable>(_ job: JobDefinition<Parameters>) {
+    public func registerJob<Parameters: JobParameters>(_ job: JobDefinition<Parameters>) {
         let builder: @Sendable (Decoder) throws -> any JobInstanceProtocol = { decoder in
             let data = try JobInstanceData<Parameters>(from: decoder)
             return try JobInstance<Parameters>(job: job, data: data)
         }
         self.builderTypeMap.withLockedValue {
-            precondition($0[job.id.name] == nil, "There is a job already registered under id \"\(job.id.name)\"")
-            $0[job.id.name] = builder
+            precondition($0[Parameters.jobName] == nil, "There is a job already registered under id \"\(Parameters.jobName)\"")
```
thoven87, line 54: Should this read "There is a job already registered under name"? or this is this fine since id is short for identifier?

Sources/Jobs/MemoryJobQueue.swift
```diff
@@ -129,18 +129,22 @@ public final class MemoryQueue: JobQueueDriver {
...
                     return nil
                 }
                 if let request = queue.popFirst() {
-                    guard request.options.delayUntil <= Date.now else {
+                    if request.options.delayUntil > Date.now {
                         self.queue.append(request)
-                        continue
+                        maxTimesToLoop -= 1
+                        if maxTimesToLoop > 0 {
+                            continue
```
thoven87, line 142: It's interesting how the continue line was not covered. Is this because the delayUntil duration is too short?

adam-fowler: Not covered? Do you mean by tests? The problem we had was the `continue` caused the handler to continually loop (if all jobs were delayed) and not give up the thread so it could be used by something else.

thoven87: Yes, I meant by test. I saw the following report from Codecov `Added line #L141 was not covered by tests`

thoven87 (approved):

PR #71, opened 2025-02-25
Add JobQueueProtocol

adam-fowler:
Similar to ApplicationProtocol

Allows us to pass around a JobQueue without caring about its driver.
I found this useful when extending the jobs example to switch between redis or Postgres depending on command line parameters


thoven87:
This is a very welcome addition. Now that this protocol has been introduced, should the queue name be part of this protocol?

thoven87 (approved):

PR #72, opened 2025-02-26
Adding Job name in metrics

### LINKED ISSUE #47: Extract job name for job metrics
It would be nice to have job name available in the metrics.

thoven87:
* Added job name to metrics
* Added job name to tracing
This PR closes #47

### DIFF
--- a/Sources/Jobs/Middleware/MetricsMiddleware.swift
+++ b/Sources/Jobs/Middleware/MetricsMiddleware.swift
@@ -73,7 +73,7 @@ public struct MetricsJobMiddleware: JobMiddleware {
         switch result {
         case .failure(let error):
             var counterDimensions: [(String, String)] = [
-                ("reason", error.code.description),
+                ("reason", error.code.description)
             ]
 
             if let jobName = error.jobName {
--- a/Sources/Jobs/Middleware/MetricsMiddleware.swift
+++ b/Sources/Jobs/Middleware/MetricsMiddleware.swift
@@ -78,7 +78,6 @@ public struct MetricsJobMiddleware: JobMiddleware {
             ]
 
             if let jobName = error.jobName {
-                
                 let jobNameDimension = ("name", jobName)
                 counterDimensions.append(jobNameDimension)
 
@@ -95,16 +94,6 @@ public struct MetricsJobMiddleware: JobMiddleware {
                 label: Self.discardedCounter,
                 dimensions: counterDimensions
             ).increment()
-
-            if let jobName = error.jobName {
-                Meter(
-                     label: Self.meterLabel,
-                     dimensions: [
-                         ("status", JobStatus.queued.rawValue),
-                         ("name", jobName),
-                     ]
-                 ).decrement()
-            }
 
         case .success(let job):
             // Decrement the current queue by 1
--- a/Sources/Jobs/Middleware/MetricsMiddleware.swift
+++ b/Sources/Jobs/Middleware/MetricsMiddleware.swift
@@ -57,7 +57,7 @@ public struct MetricsJobMiddleware: JobMiddleware {
             label: Self.meterLabel,
             dimensions: [
                 ("status", JobStatus.queued.rawValue),
-                ("name", Parameters.jobName)
+                ("name", Parameters.jobName),
             ]
         ).increment()
     }
@@ -74,7 +74,6 @@ public struct MetricsJobMiddleware: JobMiddleware {
         case .failure(let error):
             var counterDimensions: [(String, String)] = [
                 ("reason", error.code.description),
-                ("jobID", jobInstanceID),
             ]
 
             if let jobName = error.jobName {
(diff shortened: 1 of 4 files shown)


Sources/Jobs/Middleware/MetricsMiddleware.swift
```diff
@@ -148,7 +150,8 @@ public struct MetricsJobMiddleware: JobMiddleware {
                 Meter(
                     label: Self.meterLabel,
                     dimensions: [
-                        ("status", JobStatus.queued.rawValue)
+                        ("status", JobStatus.queued.rawValue),
+                        ("name", job.name),
```
adam-fowler, line 154: There is no equivalent decrement for these dimensions

Sources/Jobs/Middleware/TracingMiddleware.swift
```diff
@@ -37,6 +37,7 @@ public struct TracingJobMiddleware: JobMiddleware {
             }
             span.updateAttributes { attributes in
                 attributes["job.id"] = context.jobInstanceID
+                attributes["job.name"] = job.name
```
adam-fowler, line 40: The span is called the job name, so not sure we also need an attribute with the job name

adam-fowler (commented):
A couple of minor issues

Sources/Jobs/Middleware/MetricsMiddleware.swift
adam-fowler: With the queued status meter. There are two options. 
1) Don't include a name because it is not always available when this is edited
2) Don't decrement the meter when a name is unavailable. (Basically when a job fails to pop because it's data is so corrupt we couldn't extract a name)

What are your thoughts?

thoven87: > With the queued status meter. There are two options.
> 
> 1. Don't include a name because it is not always available when this is edited
> 2. Don't decrement the meter when a name is unavailable. (Basically when a job fails to pop because it's data is so corrupt we couldn't extract a name)
> 
> What are your thoughts?

I like this approach better. I think I implemented this approach and never pushed (I just left it in a stash). I know this off topic, but do you think it's a good idea to add observability for queue names?

thoven87: Don't decrement when name is unavailable. Since a queue still has the jobs which can't be processed.

Sources/Jobs/Middleware/MetricsMiddleware.swift
```diff
@@ -68,24 +68,27 @@ public struct MetricsJobMiddleware: JobMiddleware {
...
 
         case .success(let job):
+            // Decrement the current queue by 1
+            Meter(
+                label: Self.meterLabel,
+                dimensions: [
+                    ("status", JobStatus.queued.rawValue),
+                    ("name", job.name),
+                ]
+            ).decrement()
```
adam-fowler, line 90: You need to add the name in `onPushJobs`

Sources/Jobs/Middleware/MetricsMiddleware.swift
```diff
@@ -68,24 +68,27 @@ public struct MetricsJobMiddleware: JobMiddleware {
...
-        // Decrement the current queue by 1
-        Meter(
-            label: Self.meterLabel,
-            dimensions: [
-                ("status", JobStatus.queued.rawValue)
-            ]
-        ).decrement()
 
         switch result {
         case .failure(let error):
```
adam-fowler, line 73: Some errors contain the job name so you can use it to decrement the meter
```swift
if let jobName == error.jobName {
    Meter(
         label: Self.meterLabel,
         dimensions: [
             ("status", JobStatus.queued.rawValue),
             ("name", jobName),
         ]
     ).decrement()
}
```

adam-fowler (commented):
a couple of minor things to fix but I think we are good after that

Sources/Jobs/Middleware/MetricsMiddleware.swift
thoven87: I will take a look in two hours or so.

thoven87: We expect a job to stay in a queue (dead letter) if it fails, no?

adam-fowler: It's status gets set to failed. This will push it to the failed queue in the redis driver and set the status column in the Postgres driver I think.

It is no longer in the pending queue.

Sources/Jobs/Middleware/MetricsMiddleware.swift
```diff
@@ -68,24 +69,53 @@ public struct MetricsJobMiddleware: JobMiddleware {
...
             ).increment()
 
+            if let jobName = error.jobName {
+                Meter(
+                     label: Self.meterLabel,
+                     dimensions: [
+                         ("status", JobStatus.queued.rawValue),
+                         ("name", jobName),
+                     ]
+                 ).decrement()
```
adam-fowler, line 106: You've already run this decrement above

thoven87: Yep, it's remnant before refactoring. Good catch!

Sources/Jobs/Middleware/MetricsMiddleware.swift
```diff
@@ -68,24 +69,42 @@ public struct MetricsJobMiddleware: JobMiddleware {
...
-            dimensions: [
-                ("status", JobStatus.queued.rawValue)
-            ]
-        ).decrement()
 
         switch result {
         case .failure(let error):
+            var counterDimensions: [(String, String)] = [
+                ("reason", error.code.description),
+                ("jobID", jobInstanceID),
```
adam-fowler, line 77: Sorry one more thing. Can we remove jobID? We shouldn't have a new instance of a dimension for every instance of a job

adam-fowler (approved):

PR #73, opened 2025-02-28
Make a job’s queued at date available to its job context

thoven87:
This PR enables downstream access to when a job was queued. It affords a user to either skip the current run or process the job. This is very useful for scheduled jobs.

### DIFF
--- a/Sources/Jobs/JobContext.swift
+++ b/Sources/Jobs/JobContext.swift
@@ -14,10 +14,15 @@
 
 import Logging
 
+import struct Foundation.Date
+
 /// Context of running job
 public struct JobContext {
     /// Job instance identifier
     public let jobInstanceID: String
     /// Logger
     public let logger: Logger
+    /// The date when the job was scheduled
+    /// to run or queued
+    public let queuedAt: Date
 }
--- a/Sources/Jobs/JobParameters.swift
+++ b/Sources/Jobs/JobParameters.swift
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import struct Foundation.Date
+
 /// Defines job parameters and identifier
 public protocol JobParameters: Codable, Sendable {
     /// Job type name
--- a/Sources/Jobs/JobQueueHandler.swift
+++ b/Sources/Jobs/JobQueueHandler.swift
@@ -90,7 +90,7 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
         logger.debug("Starting Job")
         do {
             do {
-                let context = JobContext(jobInstanceID: jobID.description, logger: logger)
+                let context = JobContext(jobInstanceID: jobID.description, logger: logger, queuedAt: job.queuedAt)
                 try await self.middleware.handleJob(job: job, context: context) { job, context in
                     try await job.execute(context: context)
                 }
--- a/Tests/JobsTests/JobMiddlewareTests.swift
+++ b/Tests/JobsTests/JobMiddlewareTests.swift
@@ -78,7 +78,8 @@ final class JobMiddlewareTests: XCTestCase {
         await observers.onPopJob(result: .success(FakeJobInstance()), jobInstanceID: "0")
         XCTAssertEqual(observer1.popped, true)
         XCTAssertEqual(observer2.popped, true)
-        try await observers.handleJob(job: FakeJobInstance(), context: .init(jobInstanceID: "0", logger: .init(label: "Test"))) { _, _ in }
+        let job = FakeJobInstance()
+        try await observers.handleJob(job: job, context: .init(jobInstanceID: "0", logger: .init(label: "Test"), queuedAt: job.queuedAt)) { _, _ in }
         XCTAssertEqual(observer1.handled, true)
         XCTAssertEqual(observer2.handled, true)
     }
@@ -110,7 +111,8 @@ final class JobMiddlewareTests: XCTestCase {
             XCTAssertEqual(middleware1.pushed, first == true)
             await middlewareChain.onPopJob(result: .success(FakeJobInstance()), jobInstanceID: "0")
             XCTAssertEqual(middleware1.popped, first == true)
-            try await middlewareChain.handleJob(job: FakeJobInstance(), context: .init(jobInstanceID: "0", logger: .init(label: "Test"))) { _, _ in }
+            let job = FakeJobInstance()
+            try await middlewareChain.handleJob(job: job, context: .init(jobInstanceID: "0", logger: .init(label: "Test"), queuedAt: job.queuedAt)) { _, _ in }
             XCTAssertEqual(middleware1.handled, first == true)
         }
         try await testIf(true)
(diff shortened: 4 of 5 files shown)


thoven87:
Is there any reason why this command does not catch the swift format rules?

```sh
swift format format --parallel --in-place Sources/**/*.swift
```

thoven87:
> I'm not sure this is going to work the way you expect. The `lastScheduledAt` only gets set once ie when the scheduler starts.

hmm, I see the last scheduled date gets set [here](https://github.com/hummingbird-project/swift-jobs/pull/73/files#diff-c69ef39bca469268334bb6cde2af72296a74ab89de4ded46b43b6caed517956aR229) after a job run. It isn't the same values that's persisted in the job metadata? 

This affords a user to do the following 

```SQL
SELECT
    some_column
WHERE some_date >= \(lastScheduledAt) and some_date < \(queuedAt)
```

> 
> Anyway won't we be better to provide the `nextScheduledAt` value (This is also a value we can provide easily). eg The job handler is getting behind, such that it is going to process a job after the next scheduled version of that job. So if nextScheduledAt time is before the current time let's not bother doing it, because another job has already been scheduled.

`nextScheduledAt` is also a great idea. I didn't add it in this current implementation because it's value can be [now](https://github.com/hummingbird-project/swift-jobs/blob/main/Sources/Jobs/Scheduler/JobSchedule.swift#L61)

> So if nextScheduledAt time is before the current time let's not bother doing it, because another job has already been scheduled.

This is valid for when the scheduler is set to `latest`, but in macro batching approach where a user sets the scheduler for a given job to `.all` we can 'backfill' previous non running partitions and catch up.

adam-fowler:
You still aren't going to get what you want here. The variable lastScheduledAt is not updated, and even if it was it would be the last job that was scheduled not the last job of this type that was scheduled. I think with the `queuedAt`(which is almost exactly the same as the scheduled at time) and `nextScheduledAt` we have all the data we need.

>> So if nextScheduledAt time is before the current time let's not bother doing it, because another job has already been scheduled.

> This is valid for when the scheduler is set to latest, but in macro batching approach where a user sets the scheduler for a given job to .all we can 'backfill' previous non running partitions and catch up.

This was just an example usage.

Sources/Jobs/JobInstance.swift
```diff
@@ -60,6 +62,15 @@ extension JobInstanceProtocol {
     }
 }
 
+extension JobInstanceProtocol {
+    /// Default conformance
+    public var nextScheduledAt: Date? {
```
adam-fowler, line 67: Can we remove the default implementation

Sources/Jobs/Scheduler/JobSchedule.swift
```diff
@@ -168,7 +168,8 @@ public struct JobSchedule: MutableCollection, Sendable {
                         try await Task.sleep(until: .now + .seconds(timeInterval))
                     }
                     self.jobSchedule.updateNextScheduledDate(jobIndex: job.offset)
-                    return (job.element.jobParameters, nextScheduledDate)
+                    var nextDate: Schedule = job.element.schedule
+                    return (job.element.jobParameters, nextScheduledDate, nextDate.nextDate(after: nextScheduledDate))
```
adam-fowler, line 172: You don't need to call `nextDate.nextDate(after: nextScheduledDate)` you can get the next scheduled date from the jobSchedule as `updateNextScheduledDate` updated the `nextScheduledDate`. I think it is `self.jobSchedule[job.offset].nextScheduledDate`

The naming of the local variables is a little confusing. Maybe the original `nextScheduledDate` should be called `scheduledDate` so that we can use `nextScheduledDate` for the scheduled data after that.

Sources/Jobs/JobQueue.swift
```diff
@@ -103,6 +102,34 @@ public struct JobQueue<Queue: JobQueueDriver>: JobQueueProtocol {
...
+    /// - Returns: Identifier of queued job
+    @discardableResult internal func schedule<Parameters: JobParameters>(
+        _ parameters: Parameters,
+        currentSchedule: Date,
+        nextScheduledAt: Date?,
+        options: JobOptions
+    ) async throws -> Queue.JobID {
+        let request = JobRequest(
+            parameters: parameters,
+            queuedAt: currentSchedule,
```
adam-fowler, line 120: Should queuedAt be the value taken from the schedule or should it be `Date.now`. If the schedule processing is getting behind (Something that shouldn't happen, but you never know) we will be reporting that we queued jobs earlier than they were actually queued

Sources/Jobs/Scheduler/JobSchedule.swift
```diff
@@ -137,7 +137,7 @@ public struct JobSchedule: MutableCollection, Sendable {
 
     /// AsyncSequence of Jobs based on a JobSchedule
     struct JobSequence: AsyncSequence {
-        typealias Element = (job: JobParameters, date: Date)
+        typealias Element = (job: JobParameters, date: Date, nextScheduledAt: Date?)
```
adam-fowler, line 140: Perhaps we should convert this into a struct now, given the number of parameters is increasing
```suggestion
        struct Element {
            let job: JobParameters
            let date: Date
            let nextScheduledAt: Date?
        }
```

adam-fowler (commented):
A couple of comments. In general looks good

Sources/Jobs/Scheduler/JobSchedule.swift
thoven87: I will take a look.

Sources/Jobs/JobQueue.swift
thoven87: I used the job.date because it's the value that's stored for checking in the last time a job ran (granted, the difference  between `Date.now` and `job.date` could be milli seconds).

Sources/Jobs/Scheduler/JobSchedule.swift
thoven87: Sounds good. I will address these in the next commit.

Sources/Jobs/JobInstance.swift
thoven87: `JobInstanceProtocol` is public,  removing the default implementation will require any types conforming to said protocol to provide a value for `nextScheduledAt` even though it's only supposed to be set by the scheduler? should I make this an internal instead? since `nextScheduledAt` is supposed to be pulled from a job context.

Sources/Jobs/JobQueue.swift
adam-fowler: I suggested `Date.now` so you could avoid passing the currentSchedule value around. It appears in quite a number of function signatures

thoven87: The reason I use `job.date` is for the following scenario

We have job that is scheduled to run every hour.
Every hour, the job runs and updates the status of the system (data endpoint) with the following information:
YYYY-MM-DDTHH

For some reason the job has stopped running between hour 12 and hour 13.
After debugging for a couple of hours, we found the issue.
Pushed the fix to the repository and deployed it to production.

It's now 3 hours later. The job is now running again and the first scheduled with Date.now is for hour 15th
while job.date would have been hour 13th for the next run where accuracy is set to latest.

We could partition the data in such the "data is bucketed hourly" and re partitioned in the job itself or use the scheduler backfill the data in the expected bucket.

adam-fowler: ok fair enough

adam-fowler (approved):

PR #74, opened 2025-03-02
crontab schedule

adam-fowler:
Add support for crontab style schedules
```swift
let jobSchedule = JobSchedule([
    (job: RemoveDeadSessionsJob(), schedule: .crontab("0 4 * * 7")), // four in the morning on Sunday
    (job: BirthdayRemindersJob(), schedule: .crontab("0 9,16 * * 1-5")), // nine in the morning and four in the afternoon on week days
])
```
Syntax supported
  `*` - wildcard
  `*/n` - repeating value every so many units
  `a-b` range of values
  `a-b/n` - repeating value every so many units within range
  `a,b,c...` selection of values

If both date and week day are both set we can't support one of them being a range of values.
Currently don't support dealing with dates that go past the end of the month eg 30th Feb.  This should be fixable

### DIFF
--- a/Sources/Jobs/Scheduler/Schedule+crontab.swift
+++ b/Sources/Jobs/Scheduler/Schedule+crontab.swift
@@ -21,6 +21,11 @@ import Foundation
 #endif
 
 extension Schedule {
+    ///  Initialize Scehdule using crontab style string
+    /// - Parameters:
+    ///   - crontab: Crontab string
+    ///   - timeZone: Timezone to run schedule in
+    /// - Throws: ScheduleError for corrupt crontabs and crontabs we don't suppoty
     init(crontab: String, timeZone: TimeZone = .current) throws {
         let values = crontab.split(separator: " ", omittingEmptySubsequences: true)
         guard values.count == 5 else { throw ScheduleError("Crontab string requires 5 values") }
--- a/Sources/Jobs/Scheduler/Schedule+crontab.swift
+++ b/Sources/Jobs/Scheduler/Schedule+crontab.swift
@@ -36,7 +36,7 @@ extension Schedule {
             guard let month = Month(rawValue: $0) else { throw ScheduleError("Invalid month value") }
             return month
         }
-        let day = try Self.parse(values[4], range: 0...6) {
+        let weekDay = try Self.parse(values[4], range: 0...6) {
             guard let day = Day(rawValue: $0 + 1) else {
                 if $0 == 7 {
                     return Day.sunday
@@ -45,7 +45,7 @@ extension Schedule {
             }
             return day
         }
-        let schedule = Self(second: .specific(0), minute: minutes, hour: hours, date: date, month: month, day: day, timeZone: timeZone)
+        let schedule = Self(second: .specific(0), minute: minutes, hour: hours, date: date, month: month, day: weekDay, timeZone: timeZone)
 
         // if we have a selection set for either day or date we don't support setting the other value
         switch (schedule.day, schedule.date) {
@@ -115,7 +115,7 @@ extension Schedule {
             return .init(array)
         } else if let values = try? everyRegex.wholeMatch(in: entry) {
             let numberOfValues = (range.count + 1) / values.1
-            let array = try (0..<numberOfValues).map { try transform($0 * values.1) }
+            let array = try (0..<numberOfValues).map { try transform(range.lowerBound + $0 * values.1) }
             return .init(array)
         } else if let values = try? selectionRegex.wholeMatch(in: entry) {
             let array = try values.1.map { try transform($0) }
--- a/Sources/Jobs/Scheduler/Schedule+crontab.swift
+++ b/Sources/Jobs/Scheduler/Schedule+crontab.swift
@@ -41,6 +41,16 @@ extension Schedule {
             return day
         }
         self.init(second: .specific(0), minute: minutes, hour: hours, date: date, month: month, day: day, timeZone: timeZone)
+
+        // if we have a selection set for either day or date we don't support setting the other value
+        switch (self.day, self.date) {
+        case (.selection, .specific), (.selection, .selection), (.specific, .selection):
+            throw ScheduleError(
+                "Schedule does not support a combination of date and weekday where one is a selection of values and the other is not the wildcard '*'"
+            )
+        default:
+            break
+        }
     }
 
     static func parse<Value>(_ entry: Substring, range: ClosedRange<Int>, transform: (Int) throws -> Value) throws -> Parameter<Value> {
--- a/Sources/Jobs/Scheduler/Schedule+crontab.swift
+++ b/Sources/Jobs/Scheduler/Schedule+crontab.swift
@@ -27,6 +27,10 @@ extension Schedule {
     ///   - timeZone: Timezone to run schedule in
     /// - Throws: ScheduleError for corrupt crontabs and crontabs we don't suppoty
     static func crontab(_ crontab: String, timeZone: TimeZone = .current) throws -> Self {
+        if crontab.first == "@" {
+            guard let crontab = Self.crontabNicknames[crontab] else { throw ScheduleError("Unrecognised crontab nick name \(crontab)") }
+            return try Self.crontab(crontab, timeZone: timeZone)
+        }
         let values = crontab.split(separator: " ", omittingEmptySubsequences: true)
         guard values.count == 5 else { throw ScheduleError("Crontab string requires 5 values") }
         let minutes = try Self.parse(values[0], range: 0...59, count: 60)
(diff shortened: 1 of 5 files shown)


Sources/Jobs/Scheduler/Schedule+crontab.swift
```diff
@@ -0,0 +1,127 @@
...
+    /// - Parameters:
+    ///   - crontab: Crontab string
+    ///   - timeZone: Timezone to run schedule in
+    /// - Throws: ScheduleError for corrupt crontabs and crontabs we don't suppoty
+    static func crontab(_ crontab: String, timeZone: TimeZone = .current) throws -> Self {
+        let values = crontab.split(separator: " ", omittingEmptySubsequences: true)
+        guard values.count == 5 else { throw ScheduleError("Crontab string requires 5 values") }
+        let minutes = try Self.parse(values[0], range: 0...60) { $0 }
+        let hours = try Self.parse(values[1], range: 0...24) { $0 }
+        let date = try Self.parse(values[2], range: 0...31) { $0 }
```
thoven87, line 34: Should we keep the name as in cron syntax? i.e this variable should be called `dayOftheMonth`

Sources/Jobs/Scheduler/Schedule+crontab.swift
```diff
@@ -0,0 +1,127 @@
...
+        let month = try Self.parse(values[3], range: 1...12) {
+            guard let month = Month(rawValue: $0) else { throw ScheduleError("Invalid month value") }
+            return month
+        }
+        let day = try Self.parse(values[4], range: 0...6) {
```
thoven87, line 39: here, this would be called `dayOftheWeek`?

Sources/Jobs/Scheduler/Schedule+crontab.swift
```diff
@@ -0,0 +1,127 @@
...
+                    ","
+                }
+                OneOrMore(.digit)
+            } transform: {
+                $0.split(separator: ",").map { Int($0)! }
+            }
+            Anchor.endOfLine
+        }
+        if entry == "*" { return .any }
+        if let values = try? numberRegex.wholeMatch(in: entry) {
```
thoven87, line 111: is the exception not swallowed here?

thoven87:
This is great @adam-fowler the only thing I would are the followings

`@yearly` or `@annually`
`@monthly`
`@hourly`
`@daily` or `@midnight`
`@hourly`

I think the above presets are common enough to warrant mappings

adam-fowler:
> This is great @adam-fowler the only thing I would are the followings
> 
> 
> 
> `@yearly` or `@annually`
> 
> `@monthly`
> 
> `@hourly`
> 
> `@daily` or `@midnight`
> 
> `@hourly`
> 
> 
> 
> I think the above presets are common enough to warrant mappings 

Can't you use the functions that are already available for those. Eg `.hourly()`, `.daily(hour: 4)`

Sources/Jobs/Scheduler/Schedule.swift
```diff
@@ -291,15 +388,14 @@ public struct Schedule: Sendable {
...
     }
 
-    mutating func moveToPreviousScheduledDate() {
-        _ = self.second.prevValue()
-        _ = self.minute.prevValue()
-        _ = self.hour.prevValue()
-        _ = self.day.prevValue()
-        _ = self.date.prevValue()
-        _ = self.month.prevValue()
+    private static let daysInMonth = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31]
```
thoven87, line 394: Is there a specific reason the days are defined as above?

Tests/JobsTests/JobSchedulerTests.swift
```diff
@@ -357,6 +357,63 @@ final class JobSchedulerTests: XCTestCase {
...
+            lastScheduled: "2023-02-03T05:00:00Z",
+            now: "2023-02-03T05:29:00Z",
+            expected: ["2023-02-27T10:00:00Z", "2023-02-28T10:00:00Z", "2023-03-27T10:00:00Z", "2023-03-28T10:00:00Z"]
+        )
+        try testScheduleAfterReadingLastData(
+            schedule: .crontab("0 10 27-30 * *", timeZone: .init(secondsFromGMT: 0)!),
+            accuracy: .latest,
+            lastScheduled: "2024-02-03T05:00:00Z",
+            now: "2024-02-03T05:29:00Z",
+            expected: ["2024-02-27T10:00:00Z", "2024-02-28T10:00:00Z", "2024-02-29T10:00:00Z", "2024-03-27T10:00:00Z"]
```
thoven87, line 413: The expected values here seem a little bit funny. I would have expected, `2024-03-28T10:00:00Z`, `2024-03-29T10:00:00Z` and `2024-03-30T10:00:00Z` in the expected values as well.

adam-fowler:
I personally believe adding in multiple APIs that do the same thing is bad API design. Also the `@yearly`, `@monthly` etc do not define when in the year, month they trigger.

thoven87:
> I personally believe adding in multiple APIs that do the same thing is bad API design. Also the `@yearly`, `@monthly` etc do not define when in the year, month they trigger.

I understand your point of view Adam. It's already mutually agree that if you choose to pass in `@yearly`, it'll default to the first day of the year at midnight UTC. Same is true for `@monthly`, first day of the month at midnight UTC. If a user needs specific, the `.monthly(hour:Date)` e.t.c API can be used.

Sources/Jobs/Scheduler/Schedule.swift
adam-fowler: Not sure what you mean here. I need the day in each month so I can tell if the date has overflowed the number of days in the current month eg 31st April

Tests/JobsTests/JobSchedulerTests.swift
adam-fowler: I can add them in if you want, but the test is verifying what I wanted to verify ie we skip any out of range dates.

adam-fowler:
> I understand your point of view Adam. It's already mutually agree that if you choose to pass in `@yearly`, it'll default to the first day of the year at midnight UTC. Same is true for `@monthly`, first day of the month at midnight UTC. If a user needs specific, the `.monthly(hour:Date)` e.t.c API can be used.

I really not keen on these. But I've added the nicknames found on the linux man page (does not include midnight). If you can find a link documenting midnight I can add.

Tests/JobsTests/JobSchedulerTests.swift
thoven87: The name of the variable is what's throwing me off. I see expected, which makes me believe that's the available range from min - max and anything not defined is not expected.

Sources/Jobs/Scheduler/Schedule.swift
thoven87: I thought the Calendar API had helpers which could help here, I suppose that's not the case.

thoven87:
> > I understand your point of view Adam. It's already mutually agree that if you choose to pass in `@yearly`, it'll default to the first day of the year at midnight UTC. Same is true for `@monthly`, first day of the month at midnight UTC. If a user needs specific, the `.monthly(hour:Date)` e.t.c API can be used.
> 
> I really not keen on these. But I've added the nicknames found on the linux man page (does not include midnight). If you can find a link documenting midnight I can add.

The `0` in the syntax indicates that it's the first hour of the day which is midnight in the Gregorian calendar. Minute, hour, week are all zero based (minute is 0-59, hour is 0-23 and week is 0-6 | 0-7 where 7 is again Sunday ) If a link is still needed, I suppose either one of these should suffice 

https://en.wikipedia.org/wiki/Cron

https://man7.org/linux/man-pages/man5/crontab.5.html#top_of_page if we look at the "EXAMPLE CRON FILE" we can see this 

```
CRON_TZ=Japan
# run five minutes after midnight, every day
5 0 * * *       $HOME/bin/daily.job >> $HOME/tmp/out 2>&1
```

Sources/Jobs/Scheduler/Schedule.swift
adam-fowler: I can't find any

Tests/JobsTests/ScheduleTests.swift
```diff
@@ -182,4 +182,23 @@ final class ScheduleTests: XCTestCase {
...
+            Schedule(minute: 0, hour: 0, date: .any, month: .any, day: .any)
+        )
+        XCTAssertEqual(
+            try Schedule.crontab("@monthly"),
+            Schedule(minute: 0, hour: 0, date: 1, month: .any, day: .any)
+        )
+        XCTAssertEqual(
+            try Schedule.crontab("@yearly"),
+            Schedule(minute: 0, hour: 0, date: 1, month: .init(.january), day: .any)
+        )
```
thoven87, line 201: `@weekly` and `@annually` are missing :(

Tests/JobsTests/JobSchedulerTests.swift
thoven87: Understood! I will resolve this thread, thanks!

PR #76, opened 2025-03-05
Use FoundationEssentials when available

adam-fowler:


Sources/Jobs/JobRetry.swift
```diff
@@ -12,7 +12,23 @@
...
 //===----------------------------------------------------------------------===//
 
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
 import Foundation
+#endif
+
+#if canImport(Glibc)
+import Glibc
```
thoven87, line 22: Does this mean that we can use a [distroless images](https://github.com/GoogleContainerTools/distroless/tree/main/cc) as the final image? Also, would we need to propagate this change to the job drivers too?

adam-fowler: In effect yes, see https://www.swift.org/documentation/articles/static-linux-getting-started.html

thoven87: I see, thank you!

thoven87 (approved):

PR #77, opened 2025-03-05
Retry if driver throws an JobQueueDriverError connectionError

### LINKED ISSUE #66: Job queue iteration error handling
Currently when we call `next()` on the JobQueue iterator it can throw an error. This error is propagated further up the call stack and causes the ServiceLifecycle to instigate graceful shutdown of the application. This could be seen as a little aggressive. 

Examples of errors that can be thrown include
- Database connection failure
- Internal structure of data is not as expected. eg Postgres table not setup correctly

We could set it up so only the JobQueueHandler service is killed, but that could mean it goes unnoticed as the rest of the application is still running. Or we could ignore these errors and continue to try accessing job queue. This could mean hitting a database service repeatedly until it responds which is not very good.

We could improve the situation by defining some driver neutral errors that the job queue understands eg underlying storage connection error, structural error and have the different drivers attempt to transform their errors into something the job queue handler would recognise. eg connection errors could have an exponential backoff, structural errors cause the application to shutdown

adam-fowler:
- Add `withExponentialBackoff` which retries an operation with exponential backoff if it throws a `JobQueueDriverError` with code set to `.connectionError`.
- Wrap all driver operations in `withExponentialBackoff`
- Add driver specific retry options to `JobQueueOptions`


Sources/Jobs/JobQueueHandler.swift
```diff
@@ -54,6 +64,32 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
...
+        logger: Logger,
+        operation: () async throws -> Value
+    ) async throws -> Value {
+        var attempt = 0
+        while true {
+            do {
+                return try await operation()
+            } catch let error as JobQueueDriverError where error.code == .connectionError {
+                logger.debug("\(message()) failed")
+                if self.options.driverRetryStrategy.shouldRetry(attempt: attempt, error: error) {
```
thoven87, line 79: Should this call still be made since the default maxAttempt is set to the maximum int value? We can have a maximum of two states here where a job was popped off a queue and we loose connection to the driver and will retry until connected or the job lost connection while polling. 

For the first case, I am wondering if we should have a background running that finds jobs with states 'processing' that do not exist in a queue? Or should we by default move jobs with such state to their specific queue?

adam-fowler: So if we hit the retry limit the error is propagated further up and the job queue handler exists and we'll have to restart the queue process to continue processing jobs. The default is set to `.max` as the alternative is exiting the process. 

If the default is set to a lower number and we exit the handler then the cleanup at start can fixup any jobs left in the `processing` state.

thoven87: > So if we hit the retry limit the error is propagated further up and the job queue handler exits and we'll have to restart the queue process to continue processing jobs. The default is set to `.max` as the alternative is exiting the process.
> 
> If the default is set to a lower number and we exit the handler then the cleanup at start can fixup any jobs left in the `processing` state.

By default all the drivers are setup to do nothing on boot. I think this should be documented.

thoven87 (approved):

Sources/Jobs/JobQueueHandler.swift
adam-fowler: There is a lot of documentation to add. We have made a lot of changes since the last release

thoven87: > There is a lot of documentation to add. We have made a lot of changes since the last release

Indeed! I will help with documents too

thoven87: Also, I forgot to mention this earlier. How will this work with the Postgres driver? PostgresNIO seems to keep on retrying after a connection lost. I am that familiar with the Redis driver, I suppose it'll be same since the connection pool logic seems very similar between the two?

adam-fowler: Yeah PostgresNIO will retry connections ad-infinitum. So in theory it isn't an issue when using the Postgres driver. 

Redis is different in that it will eventually throw an error and has different errors for when an open connection was closed and when a connection couldn't be made. 

Without this change the error would be propagated up and end the job queue handler and eventually the application. 

We could move the retry to the drivers instead. I'm already asking the drivers to recognise connection errors.

adam-fowler:
I'm going to put this on hold, while I think about it. I might push this functionality down to the drivers where needed

PR #78, opened 2025-03-06
Move JobOptions to driver

adam-fowler:
- Require driver to define options for a job, using associated type JobOptions
- Add RetryJobOptions for `JobQueueDriver.retry()` which is a concrete type as the queue handler has no way to make a sensible decision on job options it doesn't know about. Drivers can make decision on what the JobOptions are based on additional knowledge it has. eg job table that includes additional data


Tests/JobsTests/JobSchedulerTests.swift
```diff
@@ -495,11 +495,11 @@ final class JobSchedulerTests: XCTestCase {
         jobSchedule.addJob(TriggerShutdownParameters(), schedule: .everyMinute(second: dateComponents.second!))
 
         // Set last date scheduled task ran as 2 seconds before scheduled job triggered
-        try await jobQueue.setMetadata(key: .jobScheduleLastDate, value: dateTriggered - 2)
+        try await jobQueue.setMetadata(key: .jobScheduleLastDate(schedulerName: "test"), value: dateTriggered - 2)
         await withThrowingTaskGroup(of: Void.self) { group in
             let serviceGroup = await ServiceGroup(
                 configuration: .init(
-                    services: [jobQueue, jobSchedule.scheduler(on: jobQueue)],
+                    services: [jobQueue, jobSchedule.scheduler(on: jobQueue, named: "test")],
```
thoven87, line 502: Would it be better if the metadata key was derived from the job queue?

```swift
jobSchedule.scheduler(on: jobQueue)
```
named could be optional if not passed in `jobQueue.name` could be used?

Sources/Jobs/Scheduler/JobSchedule.swift
```diff
@@ -93,8 +93,12 @@ public struct JobSchedule: MutableCollection, Sendable {
     ///  Create Job scheduler Service
     /// - Parameter jobQueue: Job queue to place jobs
     /// - Returns: JobScheduler
-    public func scheduler<Queue: JobQueueDriver>(on jobQueue: JobQueue<Queue>) async -> Scheduler<Queue> {
-        await .init(jobQueue: jobQueue, jobSchedule: self)
+    public func scheduler<Queue: JobQueueDriver>(
+        on jobQueue: JobQueue<Queue>,
+        named name: String = "default",
```
thoven87, line 98: as mentioned below, I think we could make name optional and derive the key from the `JobQueueDriver`. Then again, it's only ever one scheduler running and can't have a scheduler per job queue.

Tests/JobsTests/JobSchedulerTests.swift
adam-fowler: No because you might have multiple schedulers pushing to the same queue.

Sources/Jobs/Scheduler/JobSchedule.swift
adam-fowler: See comment about multiple schedulers per queue

Tests/JobsTests/JobSchedulerTests.swift
thoven87: > No because you might have multiple schedulers pushing to the same queue.

I see! I was under the impression that one would use one scheduler per job queue.

thoven87 (approved):

PR #81, opened 2025-03-06
Added JobQueueProtocol.push(_:) and test

adam-fowler:
Without the implementation without the options it is not possible to call `push` from an existential JobQueueProtocol as it has a JobOptions generic requirement. This PR resolves this by adding a version of the call without any options

### DIFF
--- a/Sources/Jobs/JobQueue.swift
+++ b/Sources/Jobs/JobQueue.swift
@@ -31,7 +31,7 @@ public protocol JobQueueProtocol: Service {
 
     ///  Push Job onto queue
     /// - Parameters:
-    ///   - id: Job identifier
+    ///   - parameters: Job parameters
     ///   - parameters: parameters for the job
     /// - Returns: Identifier of queued job
     @discardableResult func push<Parameters: JobParameters>(
@@ -49,6 +49,16 @@ public protocol JobQueueProtocol: Service {
 }
 
 extension JobQueueProtocol {
+    ///  Push Job onto queue
+    /// - Parameters:
+    ///   - parameters: Job parameters
+    /// - Returns: Identifier of queued job
+    @discardableResult public func push<Parameters: JobParameters>(
+        _ parameters: Parameters
+    ) async throws -> Queue.JobID {
+        try await self.push(parameters, options: .init())
+    }
+
     ///  Initialize JobDefinition
     /// - Parameters:
     ///   - parameters: Job parameter type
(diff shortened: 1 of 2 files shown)


Sources/Jobs/JobQueue.swift
```diff
@@ -49,6 +49,16 @@ public protocol JobQueueProtocol: Service {
 }
 
 extension JobQueueProtocol {
+    ///  Push Job onto queue
+    /// - Parameters:
+    ///   - parameters: Job parameters
+    /// - Returns: Identifier of queued job
+    @discardableResult public func push<Parameters: JobParameters>(
+        _ parameters: Parameters
```
thoven87, line 57: I think we should take the options with a default initializer and pass said option to `self.push`

adam-fowler: We already have a function that takes the parameters and the options. It's part of the protocol requirements. 

The problem is now that JobOptions is an associatedtype if you have an existential JobQueueProtocol type the compiler doesn't know what options type to generate so you can't use the function. This new function gets around  this issue by removing the options.

In general you should be using a concrete type eg `JobQueue<RedisQueue>` which is a concrete type and won't have this issue.

thoven87: The annoying part here comes in this form 

```swift
struct SomeJobQueue {
    
    struct SomeJobQueuePamameters: JobParameters {
        static var jobName: String { "SomeJobQueue" }
    }

    init(
        driver: some JobQueueDriver,
        numberOfWorkers: Int = 1,
        someService: some Service,
        mailerQueue: some JobQueueProtocol
    ) {
        queue = JobQueue(driver, numWorkers: numberOfWorkers, logger: Logger(label: "SomeJobQueue"))
        queue.registerJob(parameters: SomeJobQueuePamameters.self, maxRetryCount: 50) { parameters, context in
            // do work
            //
            context.logger("queued at \(content.queuedAt)")
            try await mailerQueue.push(
                MailerJobParameters(
                    to: "jobs@swift.jobs",
                    from: "Swift Jobs"
                ),
                options: .init() // I think we should have a default here instead of having to call init always
            )
        }
    }
}
```

adam-fowler: Please check the implementation of JobQueue.push(). It has a default argument.

thoven87 (approved):

Issue #80, opened 2025-03-06
Add queue name to JobQueueDriver protocol


adam-fowler:
Queue names are used in the drivers to differentiate between different queues. Adding the queue name to the driver protocol formulates this and allows us to use this information in the queue handler and pass it down to middleware via the JobContext

adam-fowler:
Given the issues we came across passing the queue name to the metrics and tracing middleware I'm not certain this is worthwhile.

thoven87:
> Given the issues we came across passing the queue name to the metrics and tracing middleware I'm not certain this is worthwhile.

I see! There's a limitation with the Postgres driver where one cannot have more than queue because each instance of the driver requires a migration. As a result, duplicate migrations will be found and the application will exit. Since, every features does not need to be implemented for each job driver. Queue name should be dropped from the Postgres driver and leave this feature available only to Redis.

adam-fowler:
But we have multiple queues in the Postgres driver I thought. We did this by adding a queue name column.

thoven87:
> But we have multiple queues in the Postgres driver I thought. We did this by adding a queue name column.

Yes, we did! It does not work in practice for Postgres. If we take the most "up-to-date" code from this "[POC](https://github.com/thoven87/swift-jobs-poc)" it won't work due to migration issues. `Database migration CreateSwiftJobsMigrations has been added twice.` There needs to be an option to skip duplicate migrations, the interface needs redesigned or we can just skip adding multiple queues to the Postgres driver.

PR #82, opened 2025-03-07
Add job queue name to metrics/spans

adam-fowler:
- Renamed JobContext to JobExecutionContext
- Added JobQueueContext for jobs being added/removed from queue
- Initially I added the job queue name to both of these contexts, but decided it seems a little stupid passing this values around all the time when it never changed.
- So I have gone with initialising the Metrics and Tracing middlewares with a queueName, defaulting it to "default"

### DIFF
--- a/Sources/Jobs/JobContext.swift
+++ b/Sources/Jobs/JobContext.swift
@@ -23,7 +23,7 @@ import Foundation
 /// Context of running job
 public struct JobExecutionContext {
     /// Job instance identifier
-    public let jobInstanceID: String
+    public let jobID: String
     /// Logger
     public let logger: Logger
     /// Time when the job was scheduled
@@ -36,5 +36,5 @@ public struct JobExecutionContext {
 /// context for job being adding/removed from queue
 public struct JobQueueContext {
     /// Job instance identifier
-    public let jobInstanceID: String
+    public let jobID: String
 }
--- a/Sources/Jobs/JobContext.swift
+++ b/Sources/Jobs/JobContext.swift
@@ -21,7 +21,7 @@ import Foundation
 #endif
 
 /// Context of running job
-public struct JobContext {
+public struct JobExecutionContext {
     /// Job instance identifier
     public let jobInstanceID: String
     /// Logger
@@ -31,4 +31,10 @@ public struct JobContext {
     public let queuedAt: Date
     /// Next time job is scheduled to run
     public let nextScheduledAt: Date?
+}
+
+/// context for job being adding/removed from queue
+public struct JobQueueContext {
+    /// Job instance identifier
+    public let jobInstanceID: String
 }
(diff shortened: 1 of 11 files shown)


Sources/Jobs/Middleware/MetricsMiddleware.swift
```diff
@@ -23,7 +23,12 @@ import Foundation
 
 /// Add publishing of Metrics to a job queue
 public struct MetricsJobMiddleware: JobMiddleware {
-    public init() {}
+    @usableFromInline
+    let queueName: String
+
+    public init(queueName: String = "default") {
```
thoven87, line 29: Do we expect people to pass the job queue name always? Or are we waiting for #80

Sources/Jobs/JobContext.swift
```diff
@@ -32,3 +32,9 @@ public struct JobContext {
     /// Next time job is scheduled to run
     public let nextScheduledAt: Date?
 }
+
+/// context for job being adding/removed from queue
+public struct JobQueueContext {
+    /// Job instance identifier
+    public let jobInstanceID: String
```
thoven87, line 39: should it be called `id` or `ID` since we already know this context is for a job instance?

thoven87:
Another great improvement. I just have two small questions and we're good.

Sources/Jobs/Middleware/MetricsMiddleware.swift
adam-fowler: A middleware is created without knowledge of the queue it is going to be added to. So it's going to be hard to pass the queue name from the queue.

thoven87: Once #80 is in, we should be able to the following `Queue.queueName`. We should probably do #80 first?

adam-fowler: But how do we pass the queue name to the middleware?

thoven87: > But how do we pass the queue name to the middleware?

The middleware API would need to be adjusted. `Both JobQueue` and `JobQueueHandler` already have `some JobQueueDriverProtocol`. onPush, onPush et.c can take an option which contains `jobID`, `queueName` e.t.c instead?

adam-fowler: If you read my comment in the description of this PR I decided against that as it is not very optimal sending the same data with every job.

thoven87: > If you read my comment in the description of this PR I decided against that as it is not very optimal sending the same data with every job.

I did read the comment. I suppose we can revisit this after #80 is done. It's a bit weird having to pass a queue name to the driver while also passing the same queue name in metrics middleware.

thoven87 (approved):

PR #89, opened 2025-03-12
Add optional timeout to job definition

adam-fowler:


Sources/Jobs/JobQueueHandler.swift
```diff
@@ -96,6 +96,31 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
         }
     }
 
+    /// Run job and add an optional cancellation timeout if job
+    func runJobWithOptionalTimeout(id jobID: Queue.JobID, job: any JobInstanceProtocol, logger: Logger) async throws {
```
thoven87, line 100: I am wondering how task cancellation propagation work here. Is there a possibility of having job stuck in processing state during timeout? should such event occur, we should have an internal func that handles this scenarios?

adam-fowler: No an error thrown from here will be considered a fail

Sources/Jobs/JobContext.swift
```diff
@@ -21,7 +21,7 @@ import Foundation
 #endif
 
 /// Context of running job
-public struct JobExecutionContext {
+public struct JobExecutionContext: Sendable {
     /// Job instance identifier
     public let jobID: String
```
thoven87, line 26: I always wondered if there's a specific reason why this is a string instead f a UUID?

adam-fowler: Because the jobID isn't always a UUID.

thoven87: > Because the jobID isn't always a UUID.

I think it is always `UUID`, jobName on the other hand is always a string.

Issue #85, opened 2025-03-10
Add option to cancel a job


thoven87:
As a swift jobs user, It would be nice to be able to cancel a job that's been queued if possible

adam-fowler:
I wonder if this is something that can be made driver specific as well, along with the pause

```swift
protocol JobQueueDriver {
    associatedtype Action
    
    func performAction(jobID: JobID, action: Action) async throws
}

struct PostgresQueue {
    enum Action {
        case cancel
        case pause
        case unpause
    }
    func performAction(jobID: JobID, action: Action) async throws {
    }
}
```

thoven87:
> I wonder if this is something that can be made driver specific as well, along with the pause
> 
> protocol JobQueueDriver {
>     associatedtype Action
>     
>     func performAction(jobID: JobID, action: Action) async throws
> }
> 
> struct PostgresQueue {
>     enum Action {
>         case cancel
>         case pause
>         case unpause
>     }
>     func performAction(jobID: JobID, action: Action) async throws {
>     }
> }

I think most the tickets I created will be driver specific, but I am starting to worry about the middleware because we lose observability once everything has moved to driver specific implementation.

thoven87:
> > I am starting to worry about the middleware because we lose observability once everything has moved to driver specific implementation.
> 
> Are these actions you might want to take from a middleware

Yes, and no! We should expose options to request job cancellation and resume from a user interface, but how do we surface these events through any middleware? We don't have visibility for queue name either.

thoven87:
Implemented  in [PR](#90)

PR #90, opened 2025-03-12
Resumable and cancellable jobs

### LINKED ISSUE #84: Add option to pause a job
As a swift jobs user, I would like to have an options to pause and resume jobs.

thoven87:
This PR introduces resumable and cancellable jobs and also closes #84 #85

### DIFF
--- a/Sources/Jobs/CancellableJobQueue.swift
+++ b/Sources/Jobs/CancellableJobQueue.swift
@@ -12,17 +12,13 @@
 //
 //===----------------------------------------------------------------------===//
 
-/// Protocol for Job queue driver which implements resumable jobs
-public protocol ResumeableJobQueueProtocol {
+/// Protocol for Job queue driver which implement cancellable jobs
+public protocol CancellableJobQueue {
     associatedtype JobID: CustomStringConvertible & Sendable
-    /// Pauses a job
+
+    /// Cancels a job
     /// - Parameters:
     ///  - jobID: Job id
     /// - Throws
-    func pause(jobID: JobID) async throws
-    /// Resumes a job
-    /// - Parameters:
-    ///  - jobID: Job id
-    /// - Throws
-    func resume(jobID: JobID) async throws
+    func cancel(jobID: JobID) async throws
 }
--- a/Sources/Jobs/CancellableJobQueue.swift
+++ b/Sources/Jobs/CancellableJobQueue.swift
@@ -19,6 +19,5 @@ public protocol CancellableJobQueue {
     /// Cancels a job
     /// - Parameters:
     ///  - jobID: Job id
-    /// - Throws
     func cancel(jobID: JobID) async throws
 }
--- a/Sources/Jobs/CancellableJobQueueProtocol.swift
+++ b/Sources/Jobs/CancellableJobQueueProtocol.swift
@@ -1,0 +1,24 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Hummingbird server framework project
+//
+// Copyright (c) 2025 the Hummingbird authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See hummingbird/CONTRIBUTORS.txt for the list of Hummingbird authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Protocol for Job queue driver which implement cancellable jobs
+public protocol CancellableJobQueueProtocol {
+    associatedtype JobID: CustomStringConvertible & Sendable
+
+    /// Cancels a job
+    /// - Parameters:
+    ///  - jobID: Job id
+    /// - Throws
+    func cancel(jobID: JobID) async throws
+}
--- a/Sources/Jobs/ResumeableJobQueueProtocol.swift
+++ b/Sources/Jobs/ResumeableJobQueueProtocol.swift
@@ -1,0 +1,28 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Hummingbird server framework project
+//
+// Copyright (c) 2025 the Hummingbird authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See hummingbird/CONTRIBUTORS.txt for the list of Hummingbird authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Protocol for Job queue driver which implements resumable jobs
+public protocol ResumeableJobQueueProtocol {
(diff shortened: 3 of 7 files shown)


Sources/Jobs/CancellableJobQueueProtocol.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+// See LICENSE.txt for license information
+// See hummingbird/CONTRIBUTORS.txt for the list of Hummingbird authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Protocol for Job queue driver which implement cancellable jobs
+public protocol CancellableJobQueueProtocol {
```
adam-fowler, line 16: Don't think we need Protocol at the end of the symbol name

Sources/Jobs/ResumeableJobQueueProtocol.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+// See LICENSE.txt for license information
+// See hummingbird/CONTRIBUTORS.txt for the list of Hummingbird authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Protocol for Job queue driver which implements resumable jobs
+public protocol ResumeableJobQueueProtocol {
```
adam-fowler, line 16: I think the spelling is resumable, and again I don't think we Protocol at the end of symbol name

Sources/Jobs/ResumeableJobQueueProtocol.swift
```diff
@@ -0,0 +1,28 @@
...
+    associatedtype JobID: CustomStringConvertible & Sendable
+    /// Pauses a job
+    /// - Parameters:
+    ///  - jobID: Job id
+    /// - Throws
+    func pause(jobID: JobID) async throws
+    /// Resumes a job
+    /// - Parameters:
+    ///  - jobID: Job id
+    /// - Throws
```
adam-fowler, line 26: I think you can get rid of the `/// - Throws` lines if you aren't going to detail what the function throws

Tests/JobsTests/JobsTests.swift
```diff
@@ -414,4 +414,45 @@ final class JobsTests: XCTestCase {
             XCTAssertEqual(date.advanced(by: offset).timeIntervalSinceReferenceDate, newDate.timeIntervalSinceReferenceDate, accuracy: 0.001)
         }
     }
+
+    func testPausedCancelledAndThenResume() async throws {
```
adam-fowler, line 418: Can we split this into two tests one for pause and resume and one for cancelling

adam-fowler (commented):
This all looks good, just some very minor points

Tests/JobsTests/JobsTests.swift
```diff
@@ -414,4 +414,81 @@ final class JobsTests: XCTestCase {
             XCTAssertEqual(date.advanced(by: offset).timeIntervalSinceReferenceDate, newDate.timeIntervalSinceReferenceDate, accuracy: 0.001)
         }
     }
+
+    func testCancelThenResume() async throws {
```
adam-fowler, line 418: Why are we testing resume after running a cancel? Shouldn't we just verify a cancelled job never runs. I would have thought the best way to do this would be to add two jobs, cancel the first and wait until the second job finishes. If the first job never ran you can assume it has been cancelled.

Tests/JobsTests/JobsTests.swift
```diff
@@ -414,4 +414,81 @@ final class JobsTests: XCTestCase {
...
+            let pausableJob = try await jobQueue.push(TestParameters(), options: .init(delayUntil: .now.addingTimeInterval(1)))
+            try await jobQueue.pauseJob(jobID: pausableJob)
+
+            let isEmpty = await jobQueue.queue.isEmpty()
+            XCTAssertFalse(isEmpty)
+
+            await fulfillment(of: [expectation], timeout: 5)
+            try await jobQueue.resumeJob(jobID: pausableJob)
+            let emptyAfterResume = await jobQueue.queue.isEmpty()
+            XCTAssertTrue(emptyAfterResume)
```
adam-fowler, line 490: Again I think we should we be testing for when jobs are running and not the state of the internals of the job queue. Then we can use the same test across all queues. With a job queue using one worker, run two jobs, pause the first job, wait for second job to complete, verify first job hasn't run, resume first job and wait for it to run.

Sources/Jobs/CancellableJobQueue.swift
```diff
@@ -0,0 +1,24 @@
...
+//===----------------------------------------------------------------------===//
+
+/// Protocol for Job queue driver which implement cancellable jobs
+public protocol CancellableJobQueue {
+    associatedtype JobID: CustomStringConvertible & Sendable
+
+    /// Cancels a job
+    /// - Parameters:
+    ///  - jobID: Job id
+    /// - Throws
```
adam-fowler, line 22: Can we remove the Throws

adam-fowler (commented):
Some comments on the tests

Tests/JobsTests/JobsTests.swift
thoven87: > Why are we testing resume after running a cancel? Shouldn't we just verify a cancelled job never runs. I would have thought the best way to do this would be to add two jobs, cancel the first and wait until the second job finishes. If the first job never ran you can assume it has been cancelled.

The test helper needs something similar to what's done in the Postgres job queue driver. I will update the test to have the service starts later to be able to test properly.

adam-fowler: I think you're probably right there.

thoven87: > I think you're probably right there.

I chose to use the order the jobs ran instead as it was easier for me to test.

Test sequence:

I pushed a job, paused said then job push a second job. I expect the following order: second job followed by the first job after it was resumed.

For the cancellation test, I pushed two jobs and expect only one value back.

adam-fowler (approved):
One minor question in the cancellation test, otherwise we are good

Tests/JobsTests/JobsTests.swift
```diff
@@ -414,4 +414,101 @@ final class JobsTests: XCTestCase {
...
+                configuration: .init(
+                    services: [jobQueue],
+                    gracefulShutdownSignals: [.sigterm, .sigint],
+                    logger: Logger(label: "JobQueueService")
+                )
+            )
+
+            try await jobQueue.push(TestParameters(value: 15))
+
+            let cancellable = try await jobQueue.push(TestParameters(value: 30))
```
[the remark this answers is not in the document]
thoven87: > Should the cancelled job be added before the one that runs? So we are certain it is cancelled.

I moved the cancellable job at the top.

PR #92, opened 2025-03-18
Replace single lastScheduledDate metadata with multiple

adam-fowler:
Store a last scheduled date metadata for each job. This means the state of one job schedule is not affected by the state of another. 

This fixes an issue where two jobs are scheduled at the same time but the scheduler goes down inbetween scheduling the two of them. When the scheduler is brought back up it will either add the first job into the queue again or not add the second.

### DIFF
--- a/Sources/Jobs/Scheduler/JobSchedule.swift
+++ b/Sources/Jobs/Scheduler/JobSchedule.swift
@@ -28,7 +28,7 @@ import Foundation
 ///
 /// ```swift
 /// let jobSchedule = JobSchedule([
-///     (job: RemoveDeadSessionsJob(), schedule: .weekly(day: .sunday, hour: 4))
+///     .init(job: RemoveDeadSessionsJob(), schedule: .weekly(day: .sunday, hour: 4))
 /// ])
 /// let serviceGroup = ServiceGroup(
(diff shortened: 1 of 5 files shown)


Sources/Jobs/Scheduler/JobSchedule.swift
```diff
@@ -68,6 +68,34 @@ public struct JobSchedule: MutableCollection, Sendable {
...
+                    self.nextScheduledDate = now
+                }
+            default:
+                preconditionFailure("Unsupported schedule accuracy")
+            }
+            logger.debug(
+                "First scheduled date for job",
+                metadata: [
+                    "JobName": .stringConvertible(type(of: self.jobParameters).jobName),
+                    "JobTime": .stringConvertible(self.nextScheduledDate),
```
thoven87, line 95: Should this be `NextScheduled` or `ScheduledAt`?

adam-fowler: ScheduledAt implies it has already been scheduled when it hasn't. It needs the job schedule handler to do that. This is just the value that the handler while schedule the next job at.

thoven87 (approved):
The scheduler keeps getting better and better.

PR #94, opened 2025-03-28
Add the current job attempt to the context

thoven87:
* Add current job attempt to the job context

### DIFF
--- a/Sources/Jobs/JobInstance.swift
+++ b/Sources/Jobs/JobInstance.swift
@@ -28,8 +28,8 @@ public protocol JobInstanceProtocol: Sendable {
     var retryStrategy: any JobRetryStrategy { get }
     /// Time job was queued
     var queuedAt: Date { get }
-    /// Number of attempts so far
-    var attempts: Int { get }
+    /// Current attempt
+    var attempt: Int { get }
     /// Job parameters
     var parameters: Parameters { get }
     /// Trace context
@@ -50,7 +50,7 @@ extension JobInstanceProtocol {
 
     /// Should we retry this job
     public func shouldRetry(error: Error) -> Bool {
-        self.retryStrategy.shouldRetry(attempt: self.attempts, error: error)
+        self.retryStrategy.shouldRetry(attempt: self.attempt, error: error)
     }
 
     /// Extract trace context from job instance data
@@ -77,8 +77,8 @@ struct JobInstance<Parameters: JobParameters>: JobInstanceProtocol {
     var retryStrategy: any JobRetryStrategy { job.retryStrategy }
     /// Time job was queued
     var queuedAt: Date { self.data.queuedAt }
-    /// Number of attempts so far
-    var attempts: Int { self.data.attempts }
+    /// Current attempt
+    var attempt: Int { self.data.attempt }
     /// Trace context
     var traceContext: [String: String]? { self.data.traceContext }
     /// Job parameters
@@ -104,8 +104,8 @@ public struct JobInstanceData<Parameters: JobParameters>: Codable, Sendable {
     let parameters: Parameters
     /// Time job was queued
     let queuedAt: Date
-    /// Number of attempts so far
-    let attempts: Int
+    /// Current attempt
+    let attempt: Int
     /// trace context
     let traceContext: [String: String]?
     /// Next time job is scheduled to run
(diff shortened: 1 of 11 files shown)


Tests/JobsTests/JobsTests.swift
```diff
@@ -137,6 +137,7 @@ final class JobsTests: XCTestCase {
         }
         let expectation = XCTestExpectation(description: "TestJob.execute was called", expectedFulfillmentCount: 4)
         let failedJobCount = ManagedAtomic(0)
+        let attemptCounter: NIOLockedValueBox<Int> = .init(0)
```
adam-fowler, line 140: Instead of a single value why don't you have an array which you push to and you can verify all the job attempts

thoven87: > Instead of a single value why don't you have an array which you push to and you can verify all the job attempts

Will do!

Tests/JobsTests/JobsTests.swift
```diff
@@ -135,8 +135,9 @@ final class JobsTests: XCTestCase {
         struct TestParameters: JobParameters {
             static let jobName = "testErrorRetryCount"
         }
-        let expectation = XCTestExpectation(description: "TestJob.execute was called", expectedFulfillmentCount: 4)
+        let expectation = XCTestExpectation(description: "TestJob.execute was called", expectedFulfillmentCount: 3)
```
adam-fowler, line 138: You can't just change the value in the tests. If I expect 4 attempts it should run 4 times. You need to fix the code. In JobQueueHandler call the `calculateBackoff` before incrementing the job retry value

adam-fowler: Oh you actually do, hmmm why did you need to change this value the

adam-fowler: Ah you need to change the `shouldRetry` in `ExponentialJitterJobRetryStrategy` 
```swift
    public func shouldRetry(attempt: Int, error: any Error) -> Bool {
        attempt < maxAttempts
    }
```

thoven87: > Ah you need to change the `shouldRetry` in `ExponentialJitterJobRetryStrategy`
> 
> ```swift
>     public func shouldRetry(attempt: Int, error: any Error) -> Bool {
>         attempt <= maxAttempts
>     }
> ```

I will debug [this test](https://github.com/hummingbird-project/swift-jobs/blob/main/Tests/JobsTests/JobsTests.swift#L134) because it's very confusing. We have max attempt set to 3 while the attempted values are `[1, 2, 3, 4]` I expected the job to bail after the third attempt, but that's not the case.

thoven87: I have narrowed the bug to [here](https://github.com/hummingbird-project/swift-jobs/blob/main/Sources/Jobs/JobInstance.swift#L52) `shouldRetry` is called on error. Since `attempts` is encoded as part of a job instance, it means a job might run `maxAttempt + 1`.  When `shouldRetry` we can pass the current count as `attempt + 1` this will leave the driver with the current attempts as `max - 1` or `shouldRetry` can be overloaded without the `Error` parameters and this version gets call before the `handleJob` func is called.

Doing the later will persist the current attempt count "correctly", but we also pop off a job that didn't need to be popped at all. What are your thoughts @adam-fowler

Sources/Jobs/JobInstance.swift
```diff
@@ -29,7 +29,7 @@ public protocol JobInstanceProtocol: Sendable {
     /// Time job was queued
     var queuedAt: Date { get }
     /// Number of attempts so far
-    var attempts: Int? { get }
+    var attempts: Int { get }
```
adam-fowler, line 32: Can we rename this to `attempt` as it indicates the attempt number now, and not the number of attempts that have already been made. Also edit the comment above to indicate the change in meaning.

Sources/Jobs/JobInstance.swift
```diff
@@ -105,7 +108,7 @@ public struct JobInstanceData<Parameters: JobParameters>: Codable, Sendable {
     /// Time job was queued
     let queuedAt: Date
     /// Number of attempts so far
-    let attempts: Int?
+    let attempts: Int
```
adam-fowler, line 111: Rename `attempts` to `attempt`. See above

Sources/Jobs/JobInstance.swift
```diff
@@ -114,7 +117,7 @@ public struct JobInstanceData<Parameters: JobParameters>: Codable, Sendable {
     init(
         parameters: Parameters,
         queuedAt: Date,
-        attempts: Int?,
+        attempts: Int,
```
adam-fowler, line 120: Rename `attempts` to `attempt`. See above

adam-fowler (commented):
Sorry about this @thoven87 I think I confused things. Your code prior to the last commit was actually correct. Can we just drop the last commit. I should stop attempting to review code on my phone.

I have one final minor request to reflect the change in what `attempts` means

adam-fowler (approved):

PR #99, opened 2025-04-30
Add JobQueueOptions.gracefulShutdownTimeout

adam-fowler:
If this timeout triggers after graceful shutdown has been triggered, then all the currently processing jobs are cancelled.

### DIFF
--- a/Sources/Jobs/JobQueueHandler.swift
+++ b/Sources/Jobs/JobQueueHandler.swift
@@ -58,11 +58,10 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
                     }
                 }
 
-                if let gracefulShutdownTimeout = self.options.gracefulShutdownTimeout {
-                    group.addTask {
-                        await stream.first { _ in true }
-                        try await Task.sleep(for: gracefulShutdownTimeout)
-                    }
+                group.addTask {
+                    // wait until graceful shutdown or cancellation has been triggered
+                    await stream.first { _ in true }
+                    try await Task.sleep(for: self.options.gracefulShutdownTimeout)
                 }
                 // wait on first child task to return. If the first task to return is the queue handler then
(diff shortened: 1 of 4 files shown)


Sources/Jobs/JobQueueOptions.swift
```diff
@@ -20,10 +20,21 @@ import Foundation
...
 
     /// Initialize a JobQueueOptions
-    public init(defaultRetryStrategy: any JobRetryStrategy = .exponentialJitter()) {
+    ///
+    /// - Parameters:
+    ///   - defaultRetryStrategy: Default job retry strategy for the job queue
+    ///   - gracefulShutdownTimeout: Timeout after graceful shutdown has been triggered, before jobs are cancelled
+    public init(
+        defaultRetryStrategy: any JobRetryStrategy = .exponentialJitter(),
+        gracefulShutdownTimeout: Duration? = nil
```
thoven87, line 35: Should there be a default value say 30 seconds instead of nil?

Sources/Jobs/JobQueueHandler.swift
```diff
@@ -33,30 +33,51 @@ final class JobQueueHandler<Queue: JobQueueDriver>: Sendable {
...
             }
-            await self.queue.shutdownGracefully()
         } onCancelOrGracefulShutdown: {
+            // trigger timeout
+            cont.finish()
             Task {
                 await self.queue.stop()
             }
         }
+        await self.queue.shutdownGracefully()
```
thoven87, line 80: I always find this call very funny because as far as I know only the Memory driver implements this function.

adam-fowler: It's there if a driver needs it.

Sources/Jobs/JobQueueOptions.swift
adam-fowler: Done

thoven87 (approved):

PR #100, opened 2025-05-03
Adding leader election capability

thoven87:
* Only leader should be able to schedule jobs to prevent one job from being scheduled multiple times
* The memory driver is a little bit complicated since we don't have a global state store.

### DIFF
--- a/Sources/Jobs/JobQueueDriver.swift
+++ b/Sources/Jobs/JobQueueDriver.swift
@@ -56,6 +56,8 @@ public protocol JobQueueDriver: AsyncSequence, Sendable where Element == JobQueu
     func getMetadata(_ key: String) async throws -> ByteBuffer?
     /// set job queue metadata
     func setMetadata(key: String, value: ByteBuffer) async throws
+    /// check if driver an instace is the leader
+    func isLeader() async throws -> Bool
 }
 
 extension JobQueueDriver {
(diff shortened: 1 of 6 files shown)


thoven87:
Will follow up with the Postgres driver and then Redis :(

adam-fowler:
I was wondering if it'd be simpler to write a wrapper service. 
```swift
struct LeaderService<ChildService: Service>: Service {
    let service: ChildService
    
    func run() async throws {
        // checks if current leader is still updating leadership and once it stops whoever gets in first
        // becomes leader
        try await waitUntilLeader()
        try await withThrowingTaskGroup(of: Void.self) { group in
            group.addTask {
                while true {
                    try await updateLeadership()
                    try await Task.sleep(for: .seconds(1))
            }
            group.addTask {
                try await service.run()
            }
    }
}
```

thoven87:
> I was wondering if it'd be simpler to write a wrapper service. 
> 
> ```swift
> 
> struct LeaderService<ChildService: Service>: Service {
> 
>     let service: ChildService
> 
>     
> 
>     func run() async throws {
> 
>         // checks if current leader is still updating leadership and once it stops whoever gets in first
> 
>         // becomes leader
> 
>         try await waitUntilLeader()
> 
>         try await withThrowingTaskGroup(of: Void.self) { group in
> 
>             group.addTask {
> 
>                 while true {
> 
>                     try await updateLeadership()
> 
>                     try await Task.sleep(for: .seconds(1))
> 
>                 }
> 
>             }
> 
>             group.addTask {
> 
>                 try await service.run()
> 
>             }
> 
>     }
> 
> }
> 
> ```

This implementation is decoupled from when a working is doing work. Is there a way to automatically register this service when the after the job service is ready? I also understand that if there's ever going to be one worker, we'd just end up wasting CPU cycle and bandwidth for no reason if we automatically register the Leadership service.

thoven87:
Closing this implementation. It looks like https://github.com/apple/swift-cluster-membership has the gossip protocol built-in

PR #101, opened 2025-05-15
Resurrect separation of job name and parameters

adam-fowler:
- Added JobName type which associates a job name with its parameters.
- Add functions that take JobName and Parameters which don't require Parameters to conform to JobParameters
- Remove all requirements that Parameters have to conform to `JobParameters` and replace with conformance to `Sendable & Codable`

### DIFF
--- a/Sources/Jobs/JobParameters.swift
+++ b/Sources/Jobs/JobParameters.swift
@@ -25,13 +25,6 @@ public protocol JobParameters: Codable, Sendable {
 }
 
 extension JobParameters {
-    // Job name identifier
-    public static var jobNameIdentifier: JobName<Self> {
-        .init(jobName)
-    }
-}
-
-extension JobParameters {
     /// Added so it's possible for the scheduler to add date partitions
     internal func push<Queue: JobQueueDriver>(
         to jobQueue: JobQueue<Queue>,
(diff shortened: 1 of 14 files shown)


Sources/Jobs/JobParameters.swift
```diff
@@ -24,7 +24,7 @@ public protocol JobParameters: Codable, Sendable {
     static var jobName: String { get }
 }
 
-extension JobParameters {
+/*extension JobParameters {
```
thoven87, line 27: not needed anymore?

Sources/Jobs/Scheduler/JobSchedule.swift
```diff
@@ -165,7 +204,7 @@ public struct JobSchedule: MutableCollection, Sendable {
     /// AsyncSequence of Jobs based on a JobSchedule
     struct JobSequence: AsyncSequence {
         struct Element {
-            let job: JobParameters
+            let element: JobSchedule.Element  //createJobRequest: @Sendable (Date, Date?) -> any SchedulableJobRequest
```
thoven87, line 207: same here?

thoven87 (approved):
Two minor comments

PR #104, opened 2025-06-03
Separate handler for JobQueue

adam-fowler:
They were already separate. This PR makes the separation clear in the public API

```swift
let jobQueue = JobQueue(.redis)
let serviceGroup = ServiceGroup(
    configuration: .init(
        services: [jobQueue.handler(numWorkers: 4)],
        logger: Logger(label: "JobQueueService")
    )
)
```

Also changed the initialisation setup. Previously we had an `onInit` function that was called when the job queue service was started. This never got called if the service wasn't added. This has been replaced with a `waitUntilReady` function. Currently the only job queue driver that makes use of the `onInit` was the Postgres queue and all it did was wait until the Postgres migrations were complete. This can easily be replaced with the `waitUntilReady` function.

This will fix an issue I found where the scheduler was waiting on a signal from the job queue, but this signal was only ever triggered if the job queue was added as a service.


Sources/Jobs/JobQueue.swift
```diff
@@ -44,6 +44,10 @@ public protocol JobQueueProtocol: Service {
     ///   - job: Job definition
     func registerJob(_ job: JobDefinition<some Sendable & Codable>)
 
+    ///  Create JobQueue handler that will process jobs pushed to the queue
+    /// - Parameter numWorkers: Number of workers to process jobs
+    func handler(numWorkers: Int, options: JobQueueHandlerOptions) -> any Service
```
thoven87, line 49: should it be called worker or something else? When I think of handler, it makes that this method is processing a tasks. It reality, it just offload the task to another method and await it's result?

thoven87 (approved):

Sources/Jobs/JobQueue.swift
adam-fowler: Yeah I'm not sold on the term handler. But worker is a little odd given it has a parameter numWorkers.
Perhaps "processor"?

PR #105, opened 2025-06-13
Add option to lock scheduler to one node

adam-fowler:
This requires drivers to supply a method of locking and releasing.

```swift
/// Acquire metadata lock
func acquireLock(key: String, id: ByteBuffer, expiresIn: TimeInterval) async throws -> Bool
/// Release metadata lock
func releaseLock(key: String, id: ByteBuffer) async throws
```

I also moved the wait in the JobSchedule sequence to the Scheduler


adam-fowler:
This is now ready for review. A redis implementation is here https://github.com/hummingbird-project/swift-jobs-redis/pull/42

thoven87 (approved):

PR #120, opened 2026-01-24
Add protocol for drivers that store their jobs in a queryable manner

Joannis:
It's a draft. Would like to discuss the public API prior to implementing this further.


adam-fowler (commented):
You've made job parameters the most important thing here ie you search by job parameter with a series of possible options. This assumes you know what job parameters have been registered. I think job status is probably the more interesting. When looking at the swift jobs, the first things I'd be interested in would be 
- List of pending jobs (id, name, delay)
- List of processing jobs (id, name)
- List of failed jobs (id, name)
Clicking on individual jobs would then provide the structure of the parameters. Given the parameters could be anything, it would have to be a generic tree like structure, with arrays and dictionaries and leaf nodes (string, int, double, bool etc) 

Also I think it is probably a good idea to create a wrapper type for a queryable job which holds a JobInstanceData, so we can continue to hide its contents.

adam-fowler:
Closing in favour of #144

PR #133, opened 2026-04-18
Job sleep error

adam-fowler:
Custom error that will cause JobQueueProcessor to re-queue a job at a user defined time

```swift
jobQueue.registerJob(parameters: MyJob.self) { parameters, context in
    guard let resource = getResource(parameters.resource) else {
        // resource isn't ready lets sleep for an hour
        throw JobSleep(for: .seconds(60*60))
    }
    try await processResource(resource)
}
```


thoven87 (approved):
Nice! Now, swiftjobs has "sensor" like feature using the same job id.

PR #135, opened 2026-04-21
Add JobService that combines all of the components of swift-jobs

adam-fowler:
- Added JobService type to wrap a job queue and a job schedule. It will also manage the running of both scheduler and queue processor. 
- Added `JobQueueDriver.scheduleQueueCleanup` which is called from the `JobService` to setup scheduled cleanup jobs. Added default that does nothing, but will add implementation in valkey and postgres that add regular cleanup jobs

```swift
let jobService = JobService(.memory, options: .init(processor: .init(numWorkers: 32)), logger: logger)
jobService.registerJob(parameters: TestParameters.self) { parameter, _ in
    ...
}
jobService.addScheduledJob(TestParameters(value: 56), schedule: .hourly())
async let _ = jobService.run()
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Job queue for processing workloads asynchronously across multiple nodes.
 
 ```swift
 // create job queue stored in local memory, that can run four jobs concurrently
-let jobQueue = JobQueue(
+let jobService = JobService(
     .memory, 
     logger: logger
 )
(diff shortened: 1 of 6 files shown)


thoven87 (commented):
If a user wanted to keep historical job runs using this new service. How would they define that?

adam-fowler:
> If a user wanted to keep historical job runs using this new service. How would they define that?

Do you mean keeping completed jobs? That is a driver specific option. For Postgres it would be
```swift
let jobService = JobService(
    .postgres(
        client,
        options: .init(retentionPolicy: .init(completedJobs: .retain))
    ),
    logger: logger
)
```

thoven87 (approved):

PR #144, opened 2026-07-27
API for accessing status of jobs driver

adam-fowler:
Moved into main repository.

It wasn't possible to actually implement the protocol for any of the drivers outside of their own modules. So I've decided to move this to swift-jobs and hide it behind an SPI private API.

This PR includes an implementation for the memory queue and some tests.

### DIFF
--- a/Sources/Jobs/API/JobsAPI.swift
+++ b/Sources/Jobs/API/JobsAPI.swift
@@ -1,0 +1,56 @@
+//
+// This source file is part of the Hummingbird server framework project
+// Copyright (c) the Hummingbird authors
+//
+// See LICENSE.txt for license information
+// SPDX-License-Identifier: Apache-2.0
+//
+
+public import NIOCore
+
+#if canImport(FoundationEssentials)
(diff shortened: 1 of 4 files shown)


Sources/Jobs/API/JobsAPI.swift
```diff
@@ -0,0 +1,60 @@
...
+    }
+    public let id: UUID
+    public let name: String
+    public let createdAt: Date
+    public let completedAt: Date?
+    public let status: Status
+}
+
+@_spi(JobsAPI) public struct GetJobsResponse: Sendable {
+    public let paginationToken: String?
```
thoven87, line 45: Would it be useful if there was an internal type that generate and decode pagination token? Or do we live this up to the drivers?

adam-fowler: It could be, not really certain how we'll do them for what could be a fast changing dataset.

thoven87 (approved):

Joannis (approved):

